PR #17, opened 2022-06-28
Feature: async cli upload

ryansingman:
Two major updates: asynchronous uploads and compressed uploads.

Tested locally with multiple datasets.

Benchmarking (note: compression AND asynchronicity updates are tested together):

-----------------------------------------------------
local testing -- no network condition dependence

Tweets 1M
    aiohttp (10Mb chunk): 2.68s
    sync (10Mb chunk): 3.01s

Tweets 10M
    aiohttp (1Mb chunk): 10.88s
    aiohttp (10Mb chunk): 11.48s
    sync (10Mb chunk): 29.66s

Tweets 100M
    aiohttp (1Mb chunk): 98.74s
    aiohttp (10Mb chunk): 88.63s
    sync (10Mb chunk): <gateway timeout error>

### DIFF
--- a/cleanlab_cli/api_service.py
+++ b/cleanlab_cli/api_service.py
@@ -76,9 +76,9 @@ def upload_rows(api_key, dataset_id, rows, columns):
     handle_api_error(res)
 
 
-async def upload_rows_async(session, api_key, dataset_id, rows, columns):
+async def upload_rows_async(session, api_key, dataset_id, rows, columns_json):
     url = base_url + f"/datasets/{dataset_id}"
-    data = dict(api_key=api_key, rows=json.dumps(rows), columns=json.dumps(columns))
+    data = dict(api_key=api_key, rows=json.dumps(rows), columns=columns_json)
 
     async with session.post(url=url, data=data) as res:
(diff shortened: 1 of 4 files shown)


calebchiam:
Looks like there are some CI failures because of the use of assignment expressions, but apart from that, this looks great!

My only feedback is that `upload_rows`, `upload_dataset`, and `validate_dataset` are confusingly named. Right now, `upload_rows` calls `upload_dataset`, when just going off name alone, it should be the other way around. 

Perhaps rename `upload_rows` -> `upload_dataset`, `upload_dataset` -> `upload_rows`, and `validate_dataset` -> `validate_rows`?

calebchiam (approved):

ryansingman:
Good catch on the naming changes, I'll resolve that. For the walrus operator, should I just refactor it away?

Anish has said we're targeting >=3.6 for open source and >=3.8 for studio, so what should we be targeting for the CLI? I'm assuming >=3.6 since we have it in our test matrix.

calebchiam:
Yep, >=3.6 for CLI too, just like open source, since it's still a commonly used Python version. 

I like the walrus operator, but unfortunately it must go :(

PR #20, opened 2022-07-12
Feature: upload as you iterate

ryansingman:
submit upload tasks to event loop as it iterates, rather than post iteration


calebchiam (approved):
Works well. With this setup though, processing the rows is slowed down. Before, for the 100MB (448K row) Tweets dataset, it was ~35s to validate all rows, now it's ~80s, which is about right given the 0.0001s sleep x 448K = ~45s.

This is not an issue for now, given that the bottleneck seems to lie in the backend processing and saving the uploaded rows. In the future, could investigate consuming from the upload queue more greedily or decreasing the async.io sleep time further.

PR #22, opened 2022-07-15
Redesign async upload

calebchiam:
## Description 

Async upload is currently slowed down by our use of `queue.Queue` for our `upload_queue`  -- which is needed for synchronized thread behavior, which we don't really need, since only one thread reads/writes to the upload queue at any one time.

This PR switches the data structure for `upload_queue` to `collections.deque`. Moreover, rather than processing rows one at a time, and interleaving put / get operations, we rewrite the async processes to instead have:
1. the **validate, process, put on upload queue** process fill the queue to the max of `payload_size`, before giving up control to... 
2. the **construct and POST payload** process, which empties the upload queue completely, constructs the payload from its contents using a list comprehension, sends out the POST request before yielding control.

## Motivation and Context
See this issue: https://github.com/cleanlab/cleanlab-studio/issues/228#issuecomment-1185513914 for more context.

The cost for transferring rows from queue to payload (first putting them on queue, and then removing them) is ~0.53s per payload, which adds up to ~14s for the 100MB Tweets.csv.

## How has this been tested?

This works with 10MB Tweets, but fails with 100MB Tweets at the moment. More details in the next comment.

### DIFF
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -17,3 +17,6 @@ ignore_missing_imports = True
 
 [mypy-semver.*]
 ignore_missing_imports = True
+
+[mypy-aiohttp.*]
+ignore_missing_imports = True
--- a/cleanlab_cli/dataset/upload.py
+++ b/cleanlab_cli/dataset/upload.py
@@ -1,7 +1,7 @@
 import json
 
 import click
-
+import asyncio
 from cleanlab_cli import api_service
 from cleanlab_cli import click_helpers
 from cleanlab_cli.click_helpers import progress, success, abort, info, log
@@ -17,17 +17,17 @@ from cleanlab_cli.decorators import auth_config, previous_state
 from cleanlab_cli.util import init_dataset_from_filepath
 
 
-def resume_upload(api_key, dataset_id, filepath):
+async def resume_upload(api_key, dataset_id, filepath):
     complete = api_service.get_completion_status(api_key, dataset_id)
     if complete:
         abort("Dataset is already fully uploaded.")
     saved_schema = api_service.get_dataset_schema(api_key, dataset_id)
     existing_ids = api_service.get_existing_ids(api_key, dataset_id)
-    upload_dataset(api_key, dataset_id, filepath, saved_schema, existing_ids)
+    await upload_dataset(api_key, dataset_id, filepath, saved_schema, existing_ids)
     return
 
 
-def upload_with_schema(api_key, schema, columns, filepath, prev_state):
+async def upload_with_schema(api_key, schema, columns, filepath, prev_state):
     progress("Validating provided schema...")
     loaded_schema = load_schema(schema)
     try:
@@ -43,7 +43,9 @@ def upload_with_schema(api_key, schema, columns, filepath, prev_state):
         "If this upload is interrupted, you may resume it using: cleanlab dataset upload -f"
         f" {filepath} --id {dataset_id}"
     )
-    upload_dataset(api_key=api_key, dataset_id=dataset_id, filepath=filepath, schema=loaded_schema)
+    await upload_dataset(
+        api_key=api_key, dataset_id=dataset_id, filepath=filepath, schema=loaded_schema
+    )
     return
 
 
@@ -158,7 +160,7 @@ def upload(config, prev_state, filepath, id, schema, id_column, modality, name, 
     # This is the first upload
     ## Check if uploading with schema
     if schema is not None:
-        upload_with_schema(api_key, schema, columns, filepath, prev_state)
+        asyncio.run(upload_with_schema(api_key, schema, columns, filepath, prev_state))
         return
 
     ## No schema, propose and confirm a schema
@@ -202,6 +204,8 @@ def upload(config, prev_state, filepath, id, schema, id_column, modality, name, 
         dataset_id = api_service.initialize_dataset(api_key, proposed_schema)
         info(f"Dataset initialized with ID: {dataset_id}")
         prev_state.update_args(dict(dataset_id=dataset_id))
-        upload_dataset(
-            api_key=api_key, dataset_id=dataset_id, filepath=filepath, schema=proposed_schema
+        asyncio.run(
+            upload_dataset(
+                api_key=api_key, dataset_id=dataset_id, filepath=filepath, schema=proposed_schema
+            )
         )
--- a/cleanlab_cli/dataset/upload_helpers.py
+++ b/cleanlab_cli/dataset/upload_helpers.py
@@ -265,17 +265,16 @@ async def upload_rows(
     async with aiohttp.ClientSession() as session:
         upload_tasks = []
         first_upload = True
-        import time
-
+        # import time
         # checkpoint = time.time()
         while len(upload_queue) > 0:
-            x = time.time()
+            # x = time.time()
             # print(
             #     f"time since last checkpoint {x - checkpoint}, "
             #     f"queue length: {len(upload_queue)}, "
             #     f"num_upload_tasks: {len(upload_tasks)}"
             # )
-            checkpoint = x
+            # checkpoint = x
             payload = [upload_queue.popleft() for _ in range(len(upload_queue))]
             upload_tasks.append(
                 asyncio.create_task(
--- a/cleanlab_cli/dataset/upload_helpers.py
+++ b/cleanlab_cli/dataset/upload_helpers.py
@@ -222,7 +222,7 @@ async def validate_rows(
     :param rows_per_payload: max number of rows in each payload
     :param existing_ids: set of row IDs that were already uploaded, defaults to None
     """
-    existing_ids: Set[str] = set() if existing_ids is None else set([str(x) for x in existing_ids])
+    existing_ids_: Set[str] = set() if existing_ids is None else set([str(x) for x in existing_ids])
     seen_ids: Set[str] = set()
 
     dataset = init_dataset_from_filepath(dataset_filepath)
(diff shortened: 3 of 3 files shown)


calebchiam:
Initial tests with 100MB Tweets are promising, the entire upload process takes ~45s, so we've shaved off ~15s, which is what we expected!

However, the upload fails with this error raised:
`aiohttp.client_exceptions.ClientOSError: [Errno None] Can not write request body for http://localhost:8500/api/cli/v0/datasets/63ca924594704962ada7a5f2421660c2`

Full stack trace:
```
Traceback (most recent call last):
  File "/Users/calebchiam/Documents/GitHub/cleanlab-cli/venv/lib/python3.8/site-packages/aiohttp/client_reqrep.py", line 583, in write_bytes
    await self.body.write(writer)
  File "/Users/calebchiam/Documents/GitHub/cleanlab-cli/venv/lib/python3.8/site-packages/aiohttp/payload.py", line 247, in write
    await writer.write(self._value)
  File "/Users/calebchiam/Documents/GitHub/cleanlab-cli/venv/lib/python3.8/site-packages/aiohttp/http_writer.py", line 116, in write
    self._write(chunk)
  File "/Users/calebchiam/Documents/GitHub/cleanlab-cli/venv/lib/python3.8/site-packages/aiohttp/http_writer.py", line 76, in _write
    raise ConnectionResetError("Cannot write to closing transport")
ConnectionResetError: Cannot write to closing transport

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/calebchiam/Documents/GitHub/cleanlab-cli/venv/bin/cleanlab", line 33, in <module>
    sys.exit(load_entry_point('cleanlab-cli', 'console_scripts', 'cleanlab')())
  File "/Users/calebchiam/Documents/GitHub/cleanlab-cli/venv/lib/python3.8/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/Users/calebchiam/Documents/GitHub/cleanlab-cli/venv/lib/python3.8/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/Users/calebchiam/Documents/GitHub/cleanlab-cli/venv/lib/python3.8/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/calebchiam/Documents/GitHub/cleanlab-cli/venv/lib/python3.8/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/calebchiam/Documents/GitHub/cleanlab-cli/venv/lib/python3.8/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/calebchiam/Documents/GitHub/cleanlab-cli/venv/lib/python3.8/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/Users/calebchiam/Documents/GitHub/cleanlab-cli/venv/lib/python3.8/site-packages/click/decorators.py", line 84, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/Users/calebchiam/Documents/GitHub/cleanlab-cli/venv/lib/python3.8/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/Users/calebchiam/Documents/GitHub/cleanlab-cli/venv/lib/python3.8/site-packages/click/decorators.py", line 84, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/Users/calebchiam/Documents/GitHub/cleanlab-cli/venv/lib/python3.8/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/Users/calebchiam/Documents/GitHub/cleanlab-cli/cleanlab_cli/dataset/upload.py", line 163, in upload
    asyncio.run(upload_with_schema(api_key, schema, columns, filepath, prev_state))
  File "/Users/calebchiam/opt/miniconda3/lib/python3.8/asyncio/runners.py", line 43, in run
    return loop.run_until_complete(main)
  File "/Users/calebchiam/opt/miniconda3/lib/python3.8/asyncio/base_events.py", line 616, in run_until_complete
    return future.result()
  File "/Users/calebchiam/Documents/GitHub/cleanlab-cli/cleanlab_cli/dataset/upload.py", line 46, in upload_with_schema
    await upload_dataset(
  File "/Users/calebchiam/Documents/GitHub/cleanlab-cli/cleanlab_cli/dataset/upload_helpers.py", line 361, in upload_dataset
    await asyncio.gather(producer, consumer)
  File "/Users/calebchiam/Documents/GitHub/cleanlab-cli/cleanlab_cli/dataset/upload_helpers.py", line 307, in upload_rows
    await asyncio.gather(*upload_tasks)
  File "/Users/calebchiam/Documents/GitHub/cleanlab-cli/cleanlab_cli/api_service.py", line 88, in upload_rows_async
    async with session.post(url=url, data=data, headers=headers) as res:
  File "/Users/calebchiam/Documents/GitHub/cleanlab-cli/venv/lib/python3.8/site-packages/aiohttp/client.py", line 1138, in __aenter__
    self._resp = await self._coro
  File "/Users/calebchiam/Documents/GitHub/cleanlab-cli/venv/lib/python3.8/site-packages/aiohttp/client.py", line 559, in _request
    await resp.start(conn)
  File "/Users/calebchiam/Documents/GitHub/cleanlab-cli/venv/lib/python3.8/site-packages/aiohttp/client_reqrep.py", line 898, in start
    message, payload = await protocol.read()  # type: ignore[union-attr]
  File "/Users/calebchiam/Documents/GitHub/cleanlab-cli/venv/lib/python3.8/site-packages/aiohttp/streams.py", line 616, in read
    await self._waiter
aiohttp.client_exceptions.ClientOSError: [Errno None] Can not write request body for http://localhost:8500/api/cli/v0/datasets/63ca924594704962ada7a5f2421660c2
```

anishathalye (commented):
# The bug

Here's a brief summary of what's causing the `aiohttp.client_exceptions.ClientOSError`. (It's a bug in cleanlab-cli, not in our backend or anywhere else.)

tl;dr the CPU-bound code is starving the I/O-bound code. To illustrate this, you can add the following logging:

```diff
diff --git i/cleanlab_cli/api_service.py w/cleanlab_cli/api_service.py
index 26b478f..2736941 100644
--- i/cleanlab_cli/api_service.py
+++ w/cleanlab_cli/api_service.py
@@ -68,6 +68,7 @@ def get_dataset_schema(api_key, dataset_id):
     return res.json()["schema"]


+i = 0
 async def upload_rows_async(session, api_key, dataset_id, rows, columns_json):
     url = base_url + f"/datasets/{dataset_id}"
     data = gzip.compress(
@@ -80,9 +81,15 @@ async def upload_rows_async(session, api_key, dataset_id, rows, columns_json):
         "Content-Encoding": "gzip",
     }

+    global i
+    j = i
+    i += 1
+    print(f'uploading rows, {j}')
     async with session.post(url=url, data=data, headers=headers) as res:
+        print(f'post started, entered context manager, {j}')
         res_text = await res.read()
         handle_api_error_from_json(json.loads(res_text))
+    print(f'returned from upload rows, {j}')
```

You'll see that it prints:

```
uploading rows, 0
post started, entered context manager, 0
returned from upload rows, 0
uploading rows, 1
uploading rows, 2
uploading rows, 3
uploading rows, 4
uploading rows, 5
...
```

What's happening is:

1. Producer side puts stuff in the queue, yields when the queue is full.
2. `upload_rows_async` runs until `async with session.post(url=url, data=data, headers=headers) as res:`, which effectively does an await on the async context manager's `__aenter__()`, and somewhere in there it yields back to the producer before it actualy does the HTTP POST request.
3. GOTO 1.

This means that we don't actually upload anything because we're stuck in the CPU-bound code (reading from the CSV doesn't use asyncio, so that doesn't yield to the upload tasks).

## Notes

- The reason the first payload gets through is because of the special-cased `if first_upload: await upload_tasks[0]`.
- The reason that smaller uploads work (e.g. Twitter-10M.csv) is that the CPU-bound task completes fairly quickly, at which point only the upload tasks are left, and this process happens quickly enough that the uploads can start before the connection to the server times out.
    ```
    uploading rows, 0
    post started, entered context manager, 0
    returned from upload rows, 0
    uploading rows, 1
    uploading rows, 2
    post started, entered context manager, 1
    returned from upload rows, 1
    post started, entered context manager, 2
    returned from upload rows, 2
    ```
- As an aside: if you add a print statement inside the `while len(upload_queue) >= rows_per_payload:`, you'll see that for the first payload, we `asyncio.sleep(0)` tons of times.

# Towards a fix

This can't be fixed by adding a `asyncio.sleep(0)` in the body of the loop that produces data, because then the consumer side will make tons of HTTP requests with a single row in each one. The design of this code seems a bit weird to me, where the uploader side takes whatever number of rows are available in the upload queue and uploads them, instead of doing any sort of batching. Even if this aspect were fixed, yielding at every iteration of the loop is inefficient.

At a higher level, it's a little tricky to combine I/O-bound code with CPU-bound code using asyncio without using threads/multiprocessing, because it's easy to run into these sorts of starvation issues (and there's no good solution). It might be worth having a discussion about a cleaner way to design this. Relatedly, since the CPU-bound portion is slower than the I/O-bound portion and also is parallelizable, we might be able to parallelize it as part of the same change.

If we don't want to stray too far from the current design, we could investigate using [`asyncio.loop.run_in_executor()`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor) to run the CPU-bound code in a separate thread.

PR #23, opened 2022-07-15
increase parsing speed of datetime with Timestamp

LukeMainwaring:
switching from `pd.to_datetime(column_value)` to `pd.Timestamp(column_value)` significantly speeds up processing time.

### DIFF
--- a/cleanlab_cli/dataset/upload_helpers.py
+++ b/cleanlab_cli/dataset/upload_helpers.py
@@ -128,12 +128,12 @@ def validate_and_process_record(
         else:
             if col_feature_type == "datetime":
                 try:
-                    pd.to_datetime(column_value)
+                    pd.Timestamp(column_value)
                 except (ValueError, TypeError):
                     warning = (
                         f"{column_name}: expected datetime but unable to parse '{column_value}'"
                         f" with {get_value_type(column_value)} type. Datetime strings must be"
-                        " parsable by pandas.to_datetime().",
+                        " parsable by pandas.Timestamp().",
                         ValidationWarning.TYPE_MISMATCH,
                     )
             else:


cleanlab_cli/dataset/upload_helpers.py
```diff
@@ -128,12 +128,12 @@ def validate_and_process_record(
         else:
             if col_feature_type == "datetime":
                 try:
-                    pd.to_datetime(column_value)
+                    pd.Timestamp(column_value)
                 except (ValueError, TypeError):
```
calebchiam, line 132: Do we need to change the error types here? `ValueError` is valid, not sure `TypeError` can even be raised.

calebchiam (approved):
Looks good -- though need to double-check the exception catching to make sure we don't miss a possible exception.

cleanlab_cli/dataset/upload_helpers.py
LukeMainwaring: Good point, were you thinking to narrow it or broaden it? I think we could just check for ValueError, but could change it to catch any Exception

calebchiam: Would be nice for the scope should match exactly -- but I guess there's no harm in just catching any Exception.

LukeMainwaring: Screen Shot 2022-07-15 at 5 25 46 PM

looks like TypeError is possible, but not sure if we'd every arrive at a scenario where we try to pass values other than string, int, None, etc.

calebchiam: Okay, let's leave as is then.

PR #25, opened 2022-07-19
Allow overriding API URL with an env var

anishathalye:
With this patch, rather than modifying the `api_service.py` file for local development, a developer can run the following in their shell:

```bash
export CLEANLAB_API_URL="http://localhost:8500/api/cli/v0"
```


calebchiam (approved):
LGTM! Added instructions for this in `DEVELOPMENT.md`

PR #29, opened 2022-07-23
Add badges and installation instructions

anishathalye:


calebchiam (approved):
Thanks for adding this!

PR #30, opened 2022-07-23
Add version command

anishathalye:
Having a command so users can determine the version of the installed software is really helpful for debugging users' issues.


calebchiam (approved):
Nice!

PR #32, opened 2022-07-25
float conversion for timestamp and regex for float string

LukeMainwaring:
This should improve how we handle a couple corner cases we noticed during testing.

If float is chosen as data type:

  1) convert to float before parsing with pandas.Timestamp(), 
  Ex:
  _expected datetime but unable to parse '1.413324556E+09' with string type_
  
  2) extract floating number from a string
  Ex:
  _expected 'float' but got '34%' with string type_

### DIFF
--- a/cleanlab_cli/dataset/upload_helpers.py
+++ b/cleanlab_cli/dataset/upload_helpers.py
@@ -8,6 +8,7 @@ import aiohttp
 import click
 import json
 import pandas as pd
+import re
 from decimal import Decimal
 from typing import (
     Optional,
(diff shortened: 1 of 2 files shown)


cleanlab_cli/dataset/upload_helpers.py
```diff
@@ -433,3 +443,10 @@ def save_feedback(feedback, filename):
     progress(f"Writing issues to {filename}...")
     dump_json(filename, feedback)
     success("Saved.\n")
+
+
+def extract_float_value(column_value):
```
calebchiam, line 448: Needs some docs to detail what cases this regex is able to extract float values from

cleanlab_cli/dataset/upload_helpers.py
```diff
@@ -433,3 +443,10 @@ def save_feedback(feedback, filename):
...
+    float_regex_pattern = "[-+]? (?: (?: \d* \. \d+ ) | (?: \d+ \.? ) )(?: [Ee] [+-]? \d+ ) ?"
```
calebchiam, line 449: Are there supposed to be spaces in the pattern? Tested in regex101.com and it seems like that spaces are not ignored (i.e. this pattern requires that the float contain spaces)

calebchiam (changes requested):
See above comments

cleanlab_cli/dataset/upload_helpers.py
LukeMainwaring: https://docs.python.org/3/library/re.html#simulating-scanf

This regex maps to the C library's scanf() for floating points (%e, %E, %f, %g)

calebchiam (approved):
Perfect!

PR #34, opened 2022-08-15
Put upload dataset before schema gen.

cgnorthcutt:
Re-ordering instructions to match the canonical order that most users do which is to just upload instead of generating a schema. the less common workflow should be put second.


calebchiam (approved):
Sgtm!

PR #35, opened 2022-08-23
convert to str explicitly

aayushchhabra1999:
We received a bug in production for cleanlab-studio.
image

The bug originated from
`File "/usr/local/lib/python3.10/site-packages/cleanlab_cli/dataset/schema_helpers.py", line 47, in _find_best_matching_column`

The utility expects string columns but received int columns.
This will fix the bug for now. @calebchiam - Maybe you will have more context on why the function received int columns in the first place. Let me know if there is a better solution.

**Limitation**
This will fail if the columns are not castable to string

### DIFF
--- a/cleanlab_cli/classes/excel_dataset.py
+++ b/cleanlab_cli/classes/excel_dataset.py
@@ -21,4 +21,4 @@ class ExcelDataset(Dataset):
 
     def get_columns(self) -> List[str]:
         stream = pyexcel.iget_array(file_name=self.filepath)
-        return next(stream)
+        return [str(col) for col in next(stream)]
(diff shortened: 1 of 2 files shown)


anishathalye:
I don't have full context on this, so will wait for Caleb to provide more info, but looking at the types, `columns: List[str]`, so perhaps this issue should be fixed upstream (the caller of `_find_best_matching_column` is violating the precondition / type signature of the callee).

calebchiam:
Yep, this should be fixed upstream as @anishathalye mentioned. Will tweak this PR later.

My guess (later confirmed by the Studio user who hopped on a call with me) is that the xls/xlsx filetype is encoding the first row as ints. In csv and JSON, this isn't possible, which is why it hasn't come up.

In this case, the specific user had a toy dataset without column headers. The feature and label values were both ints, hence the failure.

calebchiam (approved):
Changed this to instead ensure that the `get_columns()` method for `ExcelDataset` returns a `List[str]`, as promised by the Dataset class abstraction.

PR #37, opened 2022-09-18
Strict mypy typing and bump min Python version to 3.8

calebchiam:
- Enables strict mypy typing in CI and adds type annotations throughout `cleanlab-cli`.
- Bumps minimum Python version to 3.8, so that `typing` constructs like `Literal` and `TypedDict` can be used
- New `types.py` to hold various type annotations
- Renamed internal `feedback_log` to `warning_log`


calebchiam:
@anishathalye if you could take a quick look at `types.py` as a sanity check, that would be helpful.

I would have liked to make `ValidationWarning` an Enum, but couldn't get it to type-check given how `WarningLogType` is constructed.

PR #42, opened 2022-10-12
Tests: Unit tests for processing values based on datatypes and datetime feature type

calebchiam:
Introduces a new test suite at `tests/row_processing` that checks that dataset values are parsed / handled correctly based on the specified schema data type / feature type. 

Unit tests for the `extract_float_string` function are also introduced.

This test suite actually caught several bugs, which are fixed in this PR as well.

- Fixes bug where 'whole' floats like 1.0, 2.0, etc were not correctly cast to the corresponding integer when the data type is set to `integer` (fixes issue: https://github.com/cleanlab/cleanlab-studio/issues/574)
- Fixes bug where float strings (e.g. `'180.0'`) were replaced with `None` for `integer` data type. Now fixed so that this is converted to `180`. Note that `180.5` is still replaced by `None`.
- Fixes bug where for datetime feature types, if the value was a string and the datatype was float / int, the string was not cast to the specified data type
- Fixes bug where 1.0 and 0.0 were not recognized as boolean values. (Also includes `y` and `n` as accepted boolean strings)
- One unfixed bug: regex for `extract_float_string` is too aggressive -- alphanumeric strings like `c2ab3` result in `2` being returned instead of null

### DIFF
--- a/cleanlab_cli/dataset/upload_helpers.py
+++ b/cleanlab_cli/dataset/upload_helpers.py
@@ -221,9 +221,9 @@ def validate_and_process_record(
                 elif col_type == DataType.boolean:
                     if not isinstance(column_value, bool):
                         col_val_lower = str(column_value).lower()
-                        if col_val_lower in ["true", "t", "yes", "1"]:
+                        if col_val_lower in ["true", "t", "yes", "y", "1", "1.0"]:
                             row[column_name] = True
-                        elif col_val_lower in ["false", "f", "no", "0"]:
+                        elif col_val_lower in ["false", "f", "no", "n", "0", "0.0"]:
                             row[column_name] = False
                         else:
                             warning = (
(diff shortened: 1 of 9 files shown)


calebchiam:
Gonna merge this in so it doesn't block backend deployment.

@LukeMainwaring since you implemented the original regex, would you mind taking a look at the related test failures?

LukeMainwaring:
Yep, I've been looking into this and it's caused by the python regex `search()` operation, which searches for a match anywhere in the string. Thus, `cc982` -> `982` and `c2ab3` -> `2`. We could fix these by switching to the `match()` operation, which checks for a match only at the beginning of the string. However, this would cause issues with dollar strings. `$180.5` would return an empty string. We could switch to `match` and disallow the `$` character, or maybe check for `$` and remove it before parsing the float. Let me know your thoughts

https://docs.python.org/3/library/re.html#search-vs-match

PR #47, opened 2022-10-22
Fix regex for extracting floats/ints from strings + add more test cases

### LINKED ISSUE #33: Handle $, % in integer strings
Similar to #32, it would be helpful to include functions extract integers from dollar and percentage values.

calebchiam:
This PR fixes the regex in `extract_float_string`, which resolves the following issues:

- Negative signs were being dropped, e.g. `extract_float_string("-180")` returned `"180"` instead of `"-180"`
- Resolves #33 (we were not able to handle $ and % in strings)
- Invalid strings like `extract_float_string("c2ab3")` returned `"2"` and `extract_float_string("cc982")` returned `"982"`, instead of `""`

Adds a fuller test suite for `extract_float_string`. All tests pass now.

### DIFF
--- a/cleanlab_cli/dataset/upload_helpers.py
+++ b/cleanlab_cli/dataset/upload_helpers.py
@@ -557,14 +557,28 @@ def save_warning_log(warning_log: WarningLog, filename: str) -> None:
 
 def extract_float_string(column_value: str) -> str:
     """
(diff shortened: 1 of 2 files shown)


calebchiam:
Gonna go ahead and merge this in, but FYI @LukeMainwaring, since you [gave some helpful notes previously](https://github.com/cleanlab/cleanlab-studio/pull/42#issuecomment-1282388276)

PR #50, opened 2022-10-25
Skip checking version compat for "version" command

anishathalye:
This makes `cleanlab version` faster and also makes it avoid relying on our API. We might ask users to run `cleanlab version` to help us with debugging, and so this command should be as robust as possible.


ryansingman (approved):
LGTM

ryansingman:
(actually, looks like we need to add a type annotation for `ctx`

anishathalye:
Good catch, thanks!

PR #52, opened 2022-11-09
Fix type error

anishathalye:
We need to be careful with dict representations of things versus internal representations (like the Schema object).

This bug was preventing resuming uploads from working.


ryansingman (approved):
LGTM

PR #53, opened 2022-11-09
Parallelize S3 uploads

anishathalye:
With the Snacks dataset (848 MB, 6745 images) for example, this change brings the upload time down from 12 minutes to 1 minute.

### DIFF
--- a/cleanlab_studio/cli/api_service.py
+++ b/cleanlab_studio/cli/api_service.py
@@ -20,7 +20,7 @@ from cleanlab_studio.cli.types import JSONDict, IDType, Modality
 base_url = os.environ.get("CLEANLAB_API_BASE_URL", "https://api.cleanlab.ai/api/cli/v0")
 
 
-MAX_PARALLEL_UPLOADS = 100  # XXX choose this dynamically?
+MAX_PARALLEL_UPLOADS = 30  # XXX choose this dynamically?
 
 
 def _construct_headers(
--- a/cleanlab_studio/cli/api_service.py
+++ b/cleanlab_studio/cli/api_service.py
@@ -6,7 +6,7 @@ import gzip
 import json
 import os
 import asyncio
-from typing import List, Any, Optional
+from typing import List, Any, Optional, Tuple
 
 import aiohttp
 import requests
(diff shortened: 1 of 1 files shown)


anishathalye:
Testing with a larger dataset like MNIST: one downside is that S3 doesn't seem to like us making so many requests in parallel (parameter is currently set to 100).

cleanlab_studio/cli/api_service.py
```diff
@@ -19,6 +20,9 @@
 base_url = os.environ.get("CLEANLAB_API_BASE_URL", "https://api.cleanlab.ai/api/cli/v0")
 
 
+MAX_PARALLEL_UPLOADS = 100  # XXX choose this dynamically?
```
ryansingman, line 23: what would the optimal value for this depend on? are they things we can know before starting the upload?

I could see it making sense to do something like `min(<num_files> / <some_factor>, <max_allowed_parallel_uploads>)`

ryansingman (approved):
One comment but largely LGTM

ryansingman:
> Testing with a larger dataset like MNIST: one downside is that S3 doesn't seem to like us making so many requests in parallel (parameter is currently set to 100).

That's strange. I'm seeing this [package](https://github.com/mishudark/s3-parallel-put/blob/master/s3-parallel-put) which uses 1024 processes to upload. It [apparently performs well](https://stackoverflow.com/questions/42235618/s3-how-to-upload-large-number-of-files). It is using `boto3` though, not post requests like we're doing

anishathalye:
Some chance that the boto3 client uses a different endpoint, and AWS is applying different rate limiting rules for the two. ¯\_(ツ)_/¯ 

In any case, 30 seems to work. We can go with this and revisit later. Should probably implement backoff when getting 429 errors, retries on 500 errors, etc., which is missing from this PR.

ryansingman:
KK, that sounds good to me 👍 . Should be good to merge once you add type checking to that nested function to make mypy happy

aayushchhabra1999 (approved):
LGTM! Maybe set MAX_PARALLEL_UPLOADS = 32 as a convention (Power of 2). I actually think static parallel uploaded are more than good enough till we learn we need to make that factor dynamic to make some dataset work.

anishathalye:
Squashed and merged in e5c71a0.

PR #57, opened 2022-11-16
open editor on schema rejection

aayushchhabra1999:
Motivation:
https://github.com/cleanlab/cleanlab-studio-integration/issues/849

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -59,6 +59,7 @@ setup(
         "jsonstreams>=0.6.0",
         "semver>=2.13.0",
         "Pillow>=9.2.0",
+        "python-editor==1.0.4",
     ],
     entry_points="""
         [console_scripts]
--- a/cleanlab_studio/cli/dataset/upload.py
+++ b/cleanlab_studio/cli/dataset/upload.py
@@ -2,7 +2,7 @@ import json
 from typing import Optional, List
 
 import click
-
+import editor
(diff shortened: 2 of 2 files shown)


anishathalye (changes requested):
The core functionality works nicely! I think it does significantly improve user experience to jump straight into the editor rather than having to save the file and then edit manually.

A couple suggestions:

- For someone who isn't expecting the editor to pop up, it might be kind of jarring when the user types in the "n" and presses enter, and the editor pops up. Thoughts on how to improve this? This is a bit clunky, but for example, we could change the "Use this schema? [y/n]:" prompt to something like "Use this schema? Enter 'y' to accept, or 'n' to edit:"
- When the user enters an invalid schema, the program prints a stack trace and edits (for example, try removing the `"metadata"` key from the schema. We should instead provide a short error message (e.g., that they put in invalid JSON, or that their schema is valid JSON but an invalid schema for some reason, and explain the reason), and give them the opportunity to either edit their schema or abort.

setup.py
```diff
@@ -59,6 +59,7 @@
         "jsonstreams>=0.6.0",
         "semver>=2.13.0",
         "Pillow>=9.2.0",
+        "python-editor==1.0.4",
```
anishathalye, line 62: Leaving a note for future reference: license is Apache 2.0, which is compatible.

PR #58, opened 2022-11-16
Add basic retry logic for S3 uploads

anishathalye:
This patch adds retries with exponential backoff to S3 uploads. The implementation in this patch isn't ideal, because it only makes the failed upload back off, but all other coroutines (up to 31, with the current choice of parameters) continue uploading as before. Still, it seems to mostly alleviate the problem with failed S3 uploads due to rate limiting or transient errors.


ryansingman (commented):
Code LGTM -- testing an upload of CIFAR100 with this branch and will approve upon completion

ryansingman (approved):
CIFAR-100 worked on first try for me!

anishathalye:
Thank you!

PR #64, opened 2022-11-21
Feature: move row upload before media upload

ryansingman:
Closes https://github.com/cleanlab/cleanlab-studio-integration/issues/731


aayushchhabra1999 (commented):
What happens if the Row upload succeeds but the corresponding image upload fails?
The current design considers RDS to be the source of truth on whether a row (and thus an image) has been uploaded or not. Current design guarantees idempotence on S3 and thus things work out.

If we upload a row first and then the image upload fails, we won't be able to guarantee the presence of that image in the S3 bucket. That will break down some guarantees.

PR #67, opened 2022-11-28
Add simple image upload

anishathalye:
Implementation is a little hacky, especially with the synthesizing of the CSV metadata file and writing it to disk, but it works :) It seemed like the simplest solution without major refactoring.

Tested locally with a couple datasets. Would appreciate it if others could also test, both fresh upload + resuming an upload.

The caltech256 dataset and CIFAR10-NOISY20 dataset are formatted in a way compatible with simple image upload, but you can do other datasets too. The CreateColorsDataset tool also outputs datasets compatible with simple image upload.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,32 @@ pip install --upgrade cleanlab-studio
 
 #### Upload an image dataset
 
-An image dataset consists of a collection of image files (organized in any way, in any folder hierarchy and with any file names), along with a metadata file specifying paths and labels (and optionally, other metadata). An image dataset might be organized like this:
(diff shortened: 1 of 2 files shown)


cgnorthcutt:
example usage? `cleanlab dataset upload foldername` doesnt seem to work.

cgnorthcutt:
testing on the footwear dataset and its uploading. nice job @anishathalye !

anishathalye:
added the better error message in https://github.com/cleanlab/cleanlab-studio/pull/68

PR #69, opened 2022-12-01
check_dataset_limit for image datasets

LukeMainwaring:
Part of the solution for https://github.com/cleanlab/cleanlab-studio-integration/issues/751:

This limits limits max uploaded dataset size for image datasets, similar to how we check for text/tabular datasets. Instead of checking one dataset file size, we sum the sizes of all files in the upload. (*Value is set to 5GB for now)

Screen Shot 2022-11-30 at 7 37 37 PM

### DIFF
--- a/cleanlab_studio/cli/api_service.py
+++ b/cleanlab_studio/cli/api_service.py
@@ -241,10 +241,12 @@ def check_client_version() -> bool:
     return valid
 
 
-def check_dataset_limit(api_key: str, file_size: int, show_warning: bool = False) -> JSONDict:
+def check_dataset_limit(
+    api_key: str, file_size: int, image_dataset: bool = False, show_warning: bool = False
+) -> JSONDict:
     res = requests.post(
         base_url + "/check_dataset_limit",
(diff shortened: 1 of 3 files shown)


anishathalye (commented):
The image dataset may be missing some files, and we warn the user about those in `check_filepath_column`. `get_dataset_image_size` will crash with missing images. An easy fix might be to add a `ignore_missing_files: bool` flag to `get_file_size`, and have it ignore `FileNotFoundError`s.

LukeMainwaring:
Yes good catch, wasn't aware of that potential error. It looks like we check for both missing and unreadable filepaths - `nonexistent_filepaths = [] , unreadable_filepaths = []`. 
Should we return a list of these invalid files from `check_filepath_column` and explicitly ignore them? Or just have `get_file_size` also ignore `IOError`

anishathalye:
I think that's fine, it's unlikely to have unreadable files. Can improve this later :)

anishathalye (approved):

PR #70, opened 2022-12-19
Fix: Excel timestamp handling

ryansingman:
- convert timestamps to strings when loading ExcelDatasets

Tested with [bad.xslx](https://github.com/cleanlab/cleanlab-studio-integration/files/9675451/bad.xlsx) and [bad_numbers.xslx](https://github.com/cleanlab/cleanlab-studio-integration/files/9675408/bad_numbers.xlsx) and running:
```python
from cleanlab_studio.cli.dataset.schema_helpers import *
print(propose_schema('bad.xlsx'))
```

### DIFF
--- a/cleanlab_studio/cli/classes/excel_dataset.py
+++ b/cleanlab_studio/cli/classes/excel_dataset.py
@@ -54,7 +54,6 @@ class ExcelDataset(Dataset):
         :return: preprocessed record value
         """
         if isinstance(record_value, (datetime.time, datetime.datetime, pd.Timestamp)):
-            print(f"preprocessing datetime val: {record_value}")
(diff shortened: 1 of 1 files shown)


cleanlab_studio/version.py
```diff
@@ -1,4 +1,4 @@
-__version__ = "0.1.18"
+__version__ = "0.1.19"
```
anishathalye, line 1: Let's do version bumps separately from bug/feature PRs. Also, 0.1.19 is already released, so this will need to be in 0.1.20.

Otherwise, LGTM, feel free to merge after removing this (and feel free to bump version and tag a release after merging this).

anishathalye (approved):

PR #72, opened 2022-12-26
Feature: improved schema proposition error handling

ryansingman:
- Adds custom errors for:
  - ColumnMismatchError (number of columns in header != number of columns in data)
  - EmptyDatasetError

### DIFF
--- a/cleanlab_studio/errors.py
+++ b/cleanlab_studio/errors.py
@@ -1,10 +1,10 @@
-class InvalidDatasetError(RuntimeError):
-    ...
+class InvalidDatasetError(ValueError):
+    pass
 
 
 class EmptyDatasetError(InvalidDatasetError):
-    ...
+    pass
 
 
 class ColumnMismatchError(InvalidDatasetError):
-    ...
+    pass
--- a/cleanlab_studio/errors.py
+++ b/cleanlab_studio/errors.py
@@ -1,0 +1,10 @@
+class InvalidDatasetError(RuntimeError):
+    ...
+
+
+class EmptyDatasetError(InvalidDatasetError):
+    ...
+
+
+class ColumnMismatchError(InvalidDatasetError):
+    ...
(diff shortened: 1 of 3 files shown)


ryansingman:
Merge order:
- [CLI](https://github.com/cleanlab/cleanlab-studio/pull/72)- [Frontend](https://github.com/cleanlab/cleanlab-studio-frontend/pull/240)
- [Backend](https://github.com/cleanlab/cleanlab-studio-backend/pull/183)

cleanlab_studio/errors.py
```diff
@@ -0,0 +1,10 @@
+class InvalidDatasetError(RuntimeError):
+    ...
```
anishathalye, line 2: I think it's idiomatic to use `pass` rather than Ellipsis here

cleanlab_studio/version.py
```diff
@@ -1,4 +1,4 @@
-__version__ = "0.1.19"
+__version__ = "0.1.20"
```
anishathalye, line 1: Let's do version bumps separately

cleanlab_studio/errors.py
```diff
@@ -0,0 +1,10 @@
+class InvalidDatasetError(RuntimeError):
```
anishathalye, line 1: Subclass `ValueError` instead? It's more specific and makes sense in this context.

anishathalye (changes requested):
I like the principle of giving more specific error messages in general, though I wouldn't work too hard to support particularly badly formatted datasets like empty ones.

This change seems fine overall, aside from some minor details.

ryansingman:
> I like the principle of giving more specific error messages in general, though I wouldn't work too hard to support particularly badly formatted datasets like empty ones.
> 
> This change seems fine overall, aside from some minor details.

Agreed on not wanting to support egregiously badly formatted datasets. This change is just to allow us to catch and handle those errors properly (not resulting in unhandled exceptions in the backend)

anishathalye (approved):

PR #73, opened 2023-01-12
Hotfix: json dataset parsing

ryansingman:
Previously, parsing a JSON dataset was entirely broken -- no rows were being read. Bug can be replicated using the following dataset and the instructions below:

```python
from cleanlab_studio.cli.classes.json_dataset import JsonDataset

assert len(JsonDataset(<path to Tweets_small.json>)) == 211
```

This should raise an assertion error on main but pass on this branch.


anishathalye:
This is actually not a bug: our format asks people to upload a JSON _array_ of objects, not an id -> (rest of contents) mapping.

Screen Shot 2023-01-13 at 11 05 14 AM

PR #74, opened 2023-02-03
add relative path support to dataset record validation

axl1313:
## Description
Add a `base_directory` param to `validate_and_process_record` so that we can successfully validate datasets with relative filepaths.

## Context
We are adding support for simple image zip upload through the UI. Since we don't want to expose our internal storage organization to users, we want to write only the relative file path of an image to the dataset. The `validate_and_process_record` function is called on every row of a dataset created from the UI before saving the dataset. The validation in this function will fail if a file does not exist. We are updating it to take in a `base_directory` parameter and join that with the file path in the dataset record before checking that the file exists.

## How to Test
- Check that uploading a simple image zip dataset using these [frontend changes](https://github.com/cleanlab/cleanlab-studio-frontend/pull/277) and [backend changes](https://github.com/cleanlab/cleanlab-studio-backend/pull/236/) works properly
- Check that uploading an image dataset through the CLI still works

### DIFF
--- a/cleanlab_studio/cli/dataset/upload_helpers.py
+++ b/cleanlab_studio/cli/dataset/upload_helpers.py
@@ -3,7 +3,7 @@ Helper functions for processing and uploading dataset rows
 """
 import asyncio
 import decimal
-import os
+import pathlib
 import queue
 import re
 import threading
@@ -81,7 +81,7 @@ def validate_and_process_record(
     schema: Schema,
     seen_ids: Set[str],
     existing_ids: Set[str],
-    base_directory: str = "",
+    base_directory: pathlib.Path = pathlib.Path(""),
 ) -> Tuple[Optional[RecordType], Optional[str], Optional[RowWarningsType]]:
     """
     Validate the row against the provided schema; generate warnings where issues are found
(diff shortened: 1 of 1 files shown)


cleanlab_studio/cli/dataset/upload_helpers.py
```diff
@@ -3,6 +3,7 @@
 """
 import asyncio
 import decimal
+import os
```
ryansingman, line 6: nit: we generally try to use pathlib rather than `os` and strings for our pathing

cleanlab_studio/cli/dataset/upload_helpers.py
```diff
@@ -152,18 +154,19 @@ def validate_and_process_record(
                     )
             elif col_feature_type == FeatureType.filepath:
                 if schema.metadata.modality == Modality.image:
-                    if not image_file_exists(column_value, dataset_filepath):
+                    absolute_path = str(base_directory.joinpath(column_value))
```
ryansingman, line 157: nit: no need to convert to str

ryansingman (approved):
One nit, otherwise looks great!

PR #76, opened 2023-02-09
Refactor: create Dataset from file objects

ryansingman:
- Refactors `Dataset` class to be instantiable from file objects
- Adds dataset tests, both from filepath and from fileobj

Enables the ongoing stateless backend refactor by allowing us to initialize `Dataset`'s **without** having to save a file to the backend container filesytem.


cleanlab_studio/cli/classes/dataset.py
```diff
@@ -1,16 +1,45 @@
...
 
+    @contextlib.contextmanager
+    def fileobj(self) -> Iterator[FileObj]:
+        """Yields open IO object to dataset file."""
+        if self._filepath is not None:
+            with open(self._filepath, **self.READ_ARGS) as dataset_file:  # type: ignore
+                yield dataset_file
+
+        elif self._fileobj is not None:
+            try:
```
axl1313, line 35: I think we have to reconsider how we deal with the file object case. Since we're using `with` statements in the Dataset methods, after calling one method, the file will be closed. So calling multiple methods on the same Dataset object will fail unless the caller reopens the file in between

ryansingman: That is true when we use a context manager to create a file object -- the Dataset object created will only be valid while the file object is open. 

However, I believe we frequently have persistent file stream objects. See werkzeug.FileStorage for example -- that's how we receive filepond objects.

cleanlab_studio/cli/classes/csv_dataset.py
```diff
@@ -8,9 +8,15 @@
...
+        "mode": "r",
+        "encoding": "utf-8",
+        "errors": "ignore",
+    }
+
     def count_rows(self) -> int:
-        with open(self.filepath, "r", encoding="utf-8", errors="ignore") as f:
-            reader = csv.reader(f)
+        with self.fileobj() as f:
+            reader = csv.reader(f)  # type: ignore
```
axl1313, line 19: We might need to wrap `f` in `io.TextIOWrapper` here since it's possible that if a file object is used to initialize the dataset it was opened in byte mode

ryansingman: All CsvDatasets should have files opened in text mode. I can make that a Generic in the base class, instead of doing the weird IO[str] | IO[bytes]

cleanlab_studio/cli/util.py
```diff
@@ -65,15 +66,34 @@ def is_null_value(val: str) -> bool:
...
+    elif ext == DatasetFileExtension.json:
         return JsonDataset(filepath)
 
+    raise ValueError(f"filepath {filepath} does not have supported extension.")
+
+
+def init_dataset_from_fileobj(
+    fileobj: Union[IO[str], IO[bytes]], ext: DatasetFileExtension
+) -> Union[Dataset[IO[str]], Dataset[IO[bytes]]]:
+    """Initializes dataset from file object."""
```
ryansingman, line 84: @axl1313 do these assertions look correct to you?

axl1313: Looks mostly good to me, but might be good to make the type assertions more generic. i.e. `io.TextIOBase` instead of `io.TextIOWrapper` for the csv case so that we can support initializing a dataset from StringIO as well as TextIOWrapper. Up to you whether to make this change though since for now we know we'll only be using this function with TextIOWrapper and BytesIO

cleanlab_studio/cli/util.py
```diff
@@ -65,15 +66,34 @@ def is_null_value(val: str) -> bool:
...
+    if ext == DatasetFileExtension.csv:
+        assert isinstance(fileobj, io.TextIOBase)
+        return CsvDataset(fileobj=fileobj)
+    elif ext in [DatasetFileExtension.xls, DatasetFileExtension.xlsx]:
+        assert isinstance(fileobj, io.BytesIO)
```
axl1313, line 89: Might want to change this to `io.BufferedIOBase` so that we can support datasets initialized from open files (BufferedReader type) as well as byte streams

axl1313 (approved):
LGTM overall! Just one more minor comment

PR #77, opened 2023-02-10
add flag to disable file existence checks in validation

axl1313:
### Description
Allow disabling of filepath existence/readability checks in `validate_and_process_record`. This will be used when uploading a dataset via ZIP file since we don't extract the ZIP file contents, so the `image_file_exists` and `image_file_readable` checks will fail.

### Testing
- Unit tests pass
- Tested w/ changes in https://github.com/cleanlab/cleanlab-studio-backend/pull/273 and https://github.com/cleanlab/cleanlab-studio-frontend/pull/277

### DIFF
--- a/cleanlab_studio/cli/dataset/upload_helpers.py
+++ b/cleanlab_studio/cli/dataset/upload_helpers.py
@@ -82,7 +82,7 @@ def validate_and_process_record(
     seen_ids: Set[str],
     existing_ids: Set[str],
     base_directory: pathlib.Path = pathlib.Path(""),
-    disable_file_existence_check: bool = False,
+    disable_filepath_checks: bool = False,
 ) -> Tuple[Optional[RecordType], Optional[str], Optional[RowWarningsType]]:
     """
     Validate the row against the provided schema; generate warnings where issues are found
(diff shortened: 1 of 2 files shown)


cleanlab_studio/cli/dataset/upload_helpers.py
```diff
@@ -153,7 +153,7 @@ def validate_and_process_record(
                         ValidationWarning.TYPE_MISMATCH,
                     )
             elif col_feature_type == FeatureType.filepath:
-                if schema.metadata.modality == Modality.image:
+                if schema.metadata.modality == Modality.image and not disable_file_existence_check:
```
ryansingman, line 156: while we **do** implicitly know that the file exists for zip uploads, we don't know that the image file is readable. that check would still need to occur

axl1313: Do you think given our discussions about moving towards lazier validation, we still need to perform this check?

ryansingman: OK, fair point. Let's skip.

I would rename the flag though (we're not just disabling the file existence check, we're also disabling the image readability check)

anishathalye (approved):

PR #78, opened 2023-02-22
externally hosted media upload support

axl1313:
### Description
CLI changes to support uploading externally hosted media datasets.

### How should this be tested?
- Pull changes from `feat/external-media-upload` branch in `cleanlab-studio` and `cleanlab-studio-backend` repos
- Run studio app locally
- Try uploading dataset containing image urls from CLI (example: [dog_breeds_img_url_dataset.csv](https://github.com/cleanlab/cleanlab-studio/files/10798985/dog_breeds_img_url_dataset.csv))
- Make sure other image upload formats still work from CLI (simple image upload, csv manifest upload)
- Check if datasets were uploaded successfully in frontend (will need to pull from `feat/external-media-upload` branch in frontend repo for images to render properly when viewing project)

### DIFF
--- a/cleanlab_studio/cli/api_service.py
+++ b/cleanlab_studio/cli/api_service.py
@@ -7,16 +7,14 @@ import json
 import os
 import pathlib
 import asyncio
-from typing import IO, List, Any, Optional, Tuple, Union
+from typing import List, Any, Optional, Tuple
 
 import aiohttp
 import requests
 
 from cleanlab_studio.version import __version__
-from cleanlab_studio.cli.classes.dataset import Dataset
 from cleanlab_studio.cli.click_helpers import abort, warn, info
 from cleanlab_studio.cli.dataset.image_utils import get_image_filepath
-from cleanlab_studio.cli.dataset.schema_helpers import get_dataset_filepath_columns
 from cleanlab_studio.cli.dataset.schema_types import Schema
 from cleanlab_studio.cli.types import JSONDict, IDType, MEDIA_MODALITIES
 
--- a/cleanlab_studio/cli/api_service.py
+++ b/cleanlab_studio/cli/api_service.py
@@ -209,7 +209,7 @@ async def post_file(
     session: aiohttp.ClientSession,
     original_filepath: str,
     absolute_filepath: str,
-    post_data: Any,
+    post_data: JSONDict,
     sem: asyncio.Semaphore,
     cancelled: asyncio.Event,
 ) -> Tuple[Optional[bool], str]:
@@ -224,7 +224,6 @@ async def post_file(
     :cancelled: event that signals if upload should be cancelled
     """
     async with sem:
-        assert isinstance(post_data, dict)
         presigned_post = post_data["post"]
         # note: we pass in the path and we don't open the file until we
         # get in here, so we don't have tons of concurrently opened
--- a/cleanlab_studio/cli/api_service.py
+++ b/cleanlab_studio/cli/api_service.py
@@ -168,7 +168,7 @@ async def upload_files_for_filepath_column(
     sem = asyncio.Semaphore(MAX_PARALLEL_UPLOADS)
     cancelled = asyncio.Event()
     filepath_column_idx = columns.index(filepath_column)
-    filepaths = [row[filepath_column_idx] for row in rows]
+    filepaths: List[str] = [row[filepath_column_idx] for row in rows]
     dataset_dir: pathlib.Path = pathlib.Path(dataset_filepath).parent
     absolute_filepaths = [
         str(get_image_filepath(dataset_dir, row[filepath_column_idx])) for row in rows
@@ -186,7 +186,7 @@ async def upload_files_for_filepath_column(
                 session,
                 original_filepath,
                 absolute_filepath,
-                filepath_to_posts.get(original_filepath),
+                dict(filepath_to_posts[original_filepath]),
                 sem,
                 cancelled,
             )
--- a/cleanlab_studio/cli/api_service.py
+++ b/cleanlab_studio/cli/api_service.py
@@ -18,6 +18,7 @@ from cleanlab_studio.cli.dataset.image_utils import get_image_filepath
 from cleanlab_studio.cli.dataset.schema_helpers import get_dataset_filepath_columns
 from cleanlab_studio.cli.dataset.schema_types import Schema
 from cleanlab_studio.cli.types import JSONDict, IDType, Modality
+from cleanlab_studio.cli.util import init_dataset_from_filepath
 
 base_url = os.environ.get("CLEANLAB_API_BASE_URL", "https://api.cleanlab.ai/api/cli/v0")
 
@@ -92,7 +93,8 @@ async def upload_rows_async(
     rows: List[Any],
 ) -> None:
     modality = schema.metadata.modality
-    filepath_columns = get_dataset_filepath_columns(dataset_filepath, schema)
+    dataset = init_dataset_from_filepath(dataset_filepath)
+    filepath_columns = get_dataset_filepath_columns(dataset, schema)
     needs_media_upload = modality in [Modality.image] and filepath_columns
     columns = list(schema.fields.keys())
 
--- a/cleanlab_studio/cli/api_service.py
+++ b/cleanlab_studio/cli/api_service.py
@@ -15,6 +15,7 @@ import requests
 from cleanlab_studio.version import __version__
 from cleanlab_studio.cli.click_helpers import abort, warn, info
 from cleanlab_studio.cli.dataset.image_utils import get_image_filepath
+from cleanlab_studio.cli.dataset.schema_helpers import get_dataset_filepath_columns
 from cleanlab_studio.cli.dataset.schema_types import Schema
 from cleanlab_studio.cli.types import JSONDict, IDType, Modality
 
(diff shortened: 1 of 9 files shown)


cleanlab_studio/cli/dataset/schema_helpers.py
```diff
@@ -148,10 +150,18 @@ def multiple_separate_words_detected(values: Collection[Any]) -> bool:
 
 def is_filepath(string: str, check_existing: bool = False) -> bool:
     if check_existing:
-        return os.path.exists(string)
+        return pathlib.Path(string).exists()
     return pathlib.Path(string).suffix != "" and " " not in string
 
 
+def is_url(string: str) -> bool:
```
ryansingman, line 157: is there a reason we're not doing a `check_existing` here? seems rational to me that the default check is just seeing if it's a valid URL, and if we want to check that it exists, we can do the `head` call

cleanlab_studio/cli/dataset/schema_types.py
```diff
@@ -13,6 +13,7 @@ class DataType(Enum):
     integer = "integer"
     float = "float"
     boolean = "boolean"
+    media = "media"
```
ryansingman, line 16: why do we need a media type here? these should always be strings -- the interpretation of these strings as media is encapsulated by the `FeatureType`

cleanlab_studio/cli/dataset/schema_types.py
```diff
@@ -70,25 +71,20 @@ class SchemaMetadata:
     id_column: str
     modality: Modality
     name: str
-    filepath_column: Optional[str]
 
     @staticmethod
-    def create(
-        id_column: str, modality: str, name: str, filepath_column: Optional[str] = None
-    ) -> "SchemaMetadata":
+    def create(id_column: str, modality: str, name: str) -> "SchemaMetadata":
```
ryansingman, line 76: nit: do an `from __future__ import annotations` at the top of this file and remove the quotes around `SchemaMetadata`

ryansingman (commented):
Some comments

cleanlab_studio/cli/dataset/schema_helpers.py
axl1313: That was my initial approach, but I was having trouble figuring out how to check if the string was a valid url in a way that wouldn't include certain filepaths as valid urls (i.e. filepath starting with C:// or s3://). Do you have any ideas here?

ryansingman: Looks like this thread might have some options: https://stackoverflow.com/questions/7160737/how-to-validate-a-url-in-python-malformed-or-not

cleanlab_studio/cli/dataset/schema_types.py
axl1313: Yeah this is true. It's definitely not necessary to have a media DataType here. Mostly added it so it'd be easy to know which FeatureTypes should be interpreted as media, but there are other ways we could achieve that

ryansingman: Yeah, totally get that. I'd definitely prefer a different solution here -- it seems nice to keep the meaning of `DataType` pure

ryansingman (changes requested):
Good work here, but some iteration needed. Let me know if I can help clarify anything

cleanlab_studio/cli/dataset/schema_helpers.py
```diff
@@ -381,3 +414,19 @@ def save_schema(schema: Schema, filename: Optional[str]) -> None:
...
+    media_columns = [
+        field_name
+        for field_name, field_spec in schema.fields.items()
+        if field_spec.feature_type in MEDIA_FEATURE_TYPES
+    ]
+    df = dataset.read_file_as_dataframe()
+    return [
+        col
+        for col in media_columns
+        if not values_are_urls(df[col]) and values_are_filepaths(df[col])
```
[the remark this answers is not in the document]
axl1313: The subsampling is taken care of in the `values_are_urls`/`values_are_filepaths` functions

cleanlab_studio/cli/dataset/upload_helpers.py
```diff
@@ -683,12 +652,16 @@ def process_dataset(
 
 
 def get_image_dataset_size(
```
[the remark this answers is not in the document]
axl1313: I was thinking for URL columns, including the size of the images isn't really representative of the size of the dataset since we don't have to store the images for the user. Do we also rely on the dataset size calculated here to determine whether the dataset is under size limits for training?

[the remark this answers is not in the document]
ryansingman: So while you're right that we don't have to store the images for the user, we do care about the size of these images anyways for a few reasons:
- We use the dataset size to set a time limit for our training containers. If we don't properly account for the media size here, we won't set our training time limit correctly.
- Our required training time increases, which in turn will increase the amount we have to spend in compute. We want to properly account for this.

(A small note: I almost wrote that we incur data transfer costs, but that is not the case. AWS does not charge for in-bound data transfer, only outbound. FYI!)

There is maybe a case for handling dataset size differently for URLs, but I'm not sure that it's worth the effort. @anishathalye, do you have thoughts on this?

cleanlab_studio/cli/api_service.py
```diff
@@ -187,6 +143,117 @@ async def upload_files_for_filepath_column(filepath_column: str) -> None:
...
+        row_ids=row_ids,
+        media_type=media_type,
+    )
+    for coro in asyncio.as_completed(
+        [
+            post_file(
+                session,
+                original_filepath,
+                absolute_filepath,
+                filepath_to_posts.get(original_filepath),
```
ryansingman, line 191: we want to fail hard here if this key doesn't exist, instead of continuing with a `None` post_data

cleanlab_studio/cli/api_service.py
```diff
@@ -187,6 +143,117 @@ async def upload_files_for_filepath_column(filepath_column: str) -> None:
...
+            # coroutines
+            cancelled.set()
+            abort(f"Failed to upload {original_filepath}")
+
+
+async def post_file(
+    session: aiohttp.ClientSession,
+    original_filepath: str,
+    absolute_filepath: str,
+    post_data: Any,
```
ryansingman, line 214: this is a json dict right? let's make sure we're typing this correctly, instead of just using `Any`

cleanlab_studio/cli/dataset/schema_helpers.py
```diff
@@ -164,19 +164,21 @@ def multiple_separate_words_detected(values: Collection[Any]) -> bool:
...
+        assert isinstance(is_valid, bool)
         if check_existing:
             requests.head(string)
             return True
-        res = validators.url(string)
-        assert isinstance(res, bool)
-        return res
+        return is_valid
     except:
         return False
```
ryansingman, line 183: we don't want all of this code in the try except block I think, just the head request. It looks like we're doing something kinda funky with using an assertion for the bool-ness of the return value. I'd recomend rewriting this bit for better readability

ryansingman:
```
if validators.url(string) is not True:
    return False
    
if check_existing:
    try:
        requests.head(string)
        return True
    except requests.RequestException:
        return False
        
return True
    
```

axl1313: So the `validators.url` function is kinda wonky. It returns `True` if the url is valid and throws an error if it's not valid. That's the reason for wrapping the whole thing in a try except block. I'll rewrite to get rid of the assert though

ryansingman: based on their [docs](https://validators.readthedocs.io/en/latest/index.html#module-validators.url) and trying myself, it looks like it doesn't throw an error. it does return a `ValidationFailure` object that looks like an error however

agreed that it's a wonky API, but it shouldn't be needed to wrap that in the try except. it actually looks like the truthiness of `ValidationFailure` is false, so you can just do `if validators.url(string)...`)

axl1313: Ah yeah I just realized this haha. Assumed `ValidationFailure` was an error

ryansingman: might be good to add some unit tests to these functions -- we can do this as a refinement task but these are pretty good candidates for testing, especially edge cases here

cleanlab_studio/cli/api_service.py
axl1313: So it should be a JSONDict, but I'm calling this function by passing `filepath_to_posts.get(original_filepath)` which has `Optional[Any]` type, so type checking won't pass if we update this to JSONDict

axl1313: I can create a presigned_posts list outside of the list comprehension where we call `post_file` so that we can make sure the post exists for each filepath and add a type assertion before calling `post_file`

ryansingman: The list creation sounds like a good idea. Remember that we don't always have to do an assertion for types (which incurs a runtime cost). We can do a cast, which is essentially free computationally at runtime. In cases where we **know** what the type is, and just need to convince mypy of it, casting is typically a better solution than a type assert

cleanlab_studio/cli/api_service.py
axl1313: Do we want all the file uploads to fail if the key doesn't exist? Or just the upload for the file whose key is missing

ryansingman: we currently fail hard if any file fails to upload, so I think it should probably be the same here.

we don't have any resiliency to missing media in training afaik, so if we allowed the upload to continue with some missing data here, we'd run into issues at train time

cleanlab_studio/cli/dataset/schema_helpers.py
axl1313: I'll create a github issue for this

PR #79, opened 2023-02-22
Fix: fix simple image upload bug on windows

axl1313:
### Description
Fixes a bug with simple image upload on Windows. Tempfiles are not guaranteed to be accessible outside the original function they were created in on Windows. This led to errors when attempting to create a dataset using the temporary metadata file written for simple image datasets. This fix will use a StringIO object instead of tempfile to write the simple image metadata like we're doing for simple image zip upload through the UI.

### How to Test
- Verify that uploading a simple image dataset through CLI works
- Verify that uploading an image dataset using a manifest CSV works
- Verify that uploading text/tabular datasets still work
- Check that the above uploads work when resuming an upload that was incomplete as well

### DIFF
--- a/cleanlab_studio/cli/dataset/upload_helpers.py
+++ b/cleanlab_studio/cli/dataset/upload_helpers.py
@@ -443,7 +443,6 @@ def upload_dataset_from_stream(
     output: Optional[str] = None,
     payload_size: float = 10,
 ) -> None:
-    print(dataset_file_extension)
     dataset = init_dataset_from_fileobj(fileobj, dataset_file_extension)
     upload_dataset(
         api_key=api_key,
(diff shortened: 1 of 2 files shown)


cleanlab_studio/cli/dataset/upload_helpers.py
```diff
@@ -431,14 +433,60 @@ def check_filepath_column(modality: Modality, dataset_filepath: str, filepath_co
...
+    api_key: str,
+    dataset_id: str,
+    fileobj: Union[IO[str], IO[bytes]],
+    schema: Schema,
+    dataset_file_extension: DatasetFileExtension,
+    existing_ids: Optional[Collection[str]] = None,
+    output: Optional[str] = None,
+    payload_size: float = 10,
+) -> None:
+    print(dataset_file_extension)
```
anishathalye, line 446: debug print

anishathalye (commented):
Thank you for the quick fix!

PR #83, opened 2023-03-14
(easy) Fix: make get_dataset_file_extension case insensitive

axl1313:
Currently uploading a dataset with a capitalized file extension fails https://github.com/cleanlab/cleanlab-studio-integration/issues/872

Closes cleanlab/cleanlab-studio-integration#872


ryansingman (approved):
👍

PR #86, opened 2023-04-26
Feature: CLI using data ingestion service and support for dataframe upload through Python API

axl1313:
### Description
This PR refactors the CLI to use our new data ingestion service. It also implements support for uploading text/tabular/external media datasets from a pandas DataFrame through the Python API and support for uploading ZIP datasets using a filepath through the Python API. 

See https://github.com/cleanlab/cleanlab-studio/pull/85 for more details on CLI changes

See https://www.notion.so/cleanlab/CLI-Python-API-c5df0da937f64d7d83e5c7888ace26b9 for mostly complete Python API design

A lot of changes in this file are just moving code around. Changes to pay attention to:
- `cleanlab_studio/internal/api/api.py`
- `cleanlab_studio/internal/dataset_source/*`
- `cleanlab_studio/internal/upload_helpers.py`
- `cleanlab_studio/internal/util.py`
- `cleanlab_studio/studio/studio.py`
- `cleanlab_studio/studio/upload.py`

TODO: 
- [x] fix test failures
- [x] update README to add instructions for setting upload base url
- [x] fix type checking

Still to come:
- [ ] schema inference method for Python API
- [x] schema overrides
- [x] PySpark support
- [x] ~~support for uploading from dataframe w/o specifying dataset name~~ error message if user uploads from dataframe without dataset name
- [ ] support for specifying metadata filepath/filepath column name for ZIP w/ metadata upload

### How to Test
Checkout backend branch `feat/upload-api-w-api-key`
CLI:
- Attempt to upload text/tabular and ZIP images datasets with and without user provided schema

Python API:
- Test Pandas support:
```
from cleanlab_studio import Studio
import pandas as pd
studio = Studio("{your_api_key}")
df = {load a dataset as pandas df}
dataset_id = studio.upload_dataset(df, {dataset_name})
```
- Check that dataset was successfully uploaded in local web app
- Test PySpark support using [this notebook](https://dbc-863d8aa2-05b3.cloud.databricks.com/?o=66949918647407#notebook/3150216172252497)
    - Check that dataset is successfully uploaded in hosted dev app


cleanlab_studio/internal/api/api.py
```diff
@@ -0,0 +1,131 @@
...
+    handle_api_error(res)
+    valid: bool = res.json()["valid"]
+    return valid
+
+
+def initialize_upload(
+    api_key: str, filename: str, file_type: str, file_size: int
+) -> Tuple[str, List[int], List[str]]:
+    res = requests.get(
+        f"{upload_base_url}/initialize?size_in_bytes={file_size}&filename={filename}&file_type={file_type}",
```
ryansingman, line 64: nit: use `params` instead of f-string

cleanlab_studio/internal/dataset_source/pandas_dataset_source.py
```diff
@@ -0,0 +1,26 @@
...
+        self.name = name
+        self._fileobj = self._init_fileobj_from_df(df)
+        self.file_size = self._get_size_in_bytes()
+        self.file_type = "text/csv"
+
+    def _init_fileobj_from_df(self, df: pd.DataFrame) -> IO[bytes]:
+        fileobj = io.BytesIO()
+        df.to_csv(fileobj)
+        fileobj.seek(0)
+        return fileobj
```
ryansingman, line 21: It's worth noting that w/ Pyspark, we'll have to write to a file (by supplying a path) -- not to a stream. It may make sense to unify things a bit by having this write to a tmpfile (and doing the same for pyspark when we build that out)

cleanlab_studio/internal/dataset_source/pandas_dataset_source.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def _get_size_in_bytes(self) -> int:
+        with self.fileobj() as dataset_file:
+            dataset_file.seek(0, os.SEEK_END)
+            return dataset_file.tell()
```
ryansingman, line 26: will this not leave the cursor at the end of the file? is this an issue?

cleanlab_studio/internal/schema.py
```diff
@@ -0,0 +1,241 @@
+from dataclasses import dataclass
```
ryansingman, line 1: I don't think it makes sense to have all of these internal classes and enums here. Is there any reason why we can't keep things super simple and just pass a dict of strs?

1. I don't like that we're maintaining this both in the CLI and the backend
2. This is additional complexity that we don't want the user to touch. We should just document a format ("pass in a dictionary of field names mapped to a tuple of data type, feature type") along w/ the supported data and feature types.

ryansingman: further, we don't want to be doing _any_ validation here -- that's purely the responsibility of the studio servers

cleanlab_studio/internal/types.py
```diff
@@ -0,0 +1,19 @@
...
+JSONDict = Dict[str, Any]
+
+
+class Modality(Enum):
+    text = "text"
+    tabular = "tabular"
+    image = "image"
+
+
+MODALITIES = [m.value for m in Modality]
```
ryansingman, line 14: as above w.r.t schemas

cleanlab_studio/internal/util.py
```diff
@@ -0,0 +1,15 @@
+import pathlib
+from typing import Any, Optional
+
+import pandas as pd
+from .dataset_source import DatasetSource, FilepathDatasetSource, PandasDatasetSource
+
+
+def init_dataset_source(dataset_source: Any, name: Optional[str] = None) -> DatasetSource:
```
ryansingman, line 8: nit: use a real type for `dataset_source`

cleanlab_studio/studio/studio.py
```diff
@@ -0,0 +1,48 @@
...
+            except (FileNotFoundError, KeyError, ValueError):
+                raise ValueError(
+                    "No API key found; either specify API key or log in with 'cleanlab login' first"
+                )
+        api.validate_api_key(api_key)
+        self._api_key = api_key
+
+    def upload_dataset(
+        self,
+        dataset: Any,  # for now, a spark.DataFrame; in the future, we will support more
```
ryansingman, line 30: comment is incorrect

ryansingman (commented):
Changes largely LGTM, with the exception of the schema stuff. Lemme know if you wanna chat more about that if my comment wasn't clear / you disagree

Also, can you remove tests that are no longer applicable?

cleanlab_studio/internal/dataset_source/pandas_dataset_source.py
axl1313: Oops yeah I meant to seek back to 0 before returning. Good catch!

cleanlab_studio/internal/schema.py
axl1313: Can't immediately think of any reasons against just using a dictionary instead of the Schema object here. Definitely agree that it's bad to have all these definitions duplicated in CLI and data ingestion. There are some backend dependencies on the version defined in CLI right now though, so if we remove this, we'll have to update those as well

cleanlab_studio/internal/dataset_source/pandas_dataset_source.py
axl1313: Actually this is also fine as is because the `self.fileobj()` context manager contains a `finally: fileobj.seek(0)`

ryansingman: got it, didn't see that! thx

cleanlab_studio/internal/schema.py
axl1313: Most of the usage should go away with the data ingestion refactor. We're currently using the Schema object defined in CLI to [create simple zip schema](https://github.com/cleanlab/cleanlab-studio-backend/blob/982c28576668041334bfdaacf3745ebe53ad8a53/backend/api/upload/upload_api_helpers.py#L20) after the data ingestion refactor, but we could move the simple zip schema creation inside the data ingestion service instead. 

Other than that there's some references to CLI types in [training_service](https://github.com/cleanlab/cleanlab-studio-backend/blob/b56791f0dbef51ca1cbe144a96102bd9245ae0af/backend/services/training_service/training_service.py#L7) and [media_upload_service](https://github.com/cleanlab/cleanlab-studio-backend/blob/c2fc1c295419f4babde4f51d6a58c1b32649e5cc/backend/services/media_upload_service/media_upload_service.py)

ryansingman: Got it, ok. Let's take this time to (finally) create a Cleanlab-studio-lib repo to hold this stuff. I'll do so this afternoon, and then we can remove this code from the CLI

cleanlab_studio/internal/dataset_source/pandas_dataset_source.py
axl1313: Note to self: test on windows after rewriting this to use tempfile (we had issues with tempfiles on windows not behaving the same way previously)

cleanlab_studio/cli/dataset/schema.py
```diff
@@ -1,4 +1,3 @@
-import pathlib
```
ryansingman, line 1: I'd like to deprecate this module altogether. We are moving schema validation to the backend, so users shouldn't be generating/validating schemas at all

ryansingman (commented):
Changes look great! Some general notes before testing:
- We're going to want to do a major version bump for this
- We need to add documentation for the python API. Can you speak w/ @anishathalye on this?

ryansingman:
Thoughts from testing CLI:
- Seems to work as intended! Nice work :)
- Need to add a progress bar for the upload file stage -- otherwise the CLI seems non-responsive for large dataset uploads
- nit: can you float format the progress bars? I only saw this specifically for ingesting (though I assume it applies to both), but we don't need like 20 sig figs to indicate upload progress 😆

cleanlab_studio/internal/util.py
```diff
@@ -13,16 +12,18 @@
...
 
 
 def init_dataset_source(
     dataset_source: DatasetSourceType, dataset_name: Optional[str] = None
 ) -> DatasetSource:
     if isinstance(dataset_source, pd.DataFrame):
+        assert dataset_name is not None
         return PandasDatasetSource(df=dataset_source, dataset_name=dataset_name)
     elif isinstance(dataset_source, pyspark.sql.DataFrame):
+        assert dataset_name is not None
```
ryansingman: this is a user-facing error, so let's give a good error message. smthn like:

`dataset_name` must be provided for `dataset_source` of type `(pd|pyspark.sql).DataFrame`

cleanlab_studio/studio/studio.py
```diff
@@ -34,6 +35,8 @@ def upload_dataset(
         id_column: Optional[str] = None,
     ) -> str:
         ds = init_dataset_source(dataset, dataset_name)
+        if isinstance(ds, DataFrameDatasetSource) and dataset_name is None:
+            raise ValueError("Must provide dataset name if uploading from a DataFrame")
```
ryansingman, line 39: won't we already have hit the assertion above?

setup.py
```diff
@@ -55,7 +55,7 @@
         "tqdm>=4.64.0",
         "ijson>=3.1.4",
         "jsonstreams>=0.6.0",
-        "semver>=2.13.0",
+        "semver>=2.13.0,<3.0.0",
```
ryansingman, line 58: reason for this change?

cleanlab_studio/studio/upload.py
```diff
@@ -17,16 +17,20 @@ def upload_dataset(
...
     schema = get_proposed_schema(api_key, upload_id)
-    if schema_overrides is not None:
-        for field in schema_overrides:
-            schema["fields"][field] = schema_overrides[field]
-    if modality is not None:
-        schema["metadata"]["modality"] = modality
-    if id_column is not None:
-        schema["metadata"]["id_column"] = id_column
+
+    # if not simple zip upload, apply schema overrides
```
ryansingman, line 24: can we add an error message for the case where a user does supply schema overrides to a simple zip upload?

cleanlab_studio/studio/studio.py
axl1313: Oh yes good catch, I'll move this error message inside `init_dataset_source`

setup.py
axl1313: There were mypy errors with v3.0.0 of semver not explicitly exporting compare and VersionInfo

anishathalye (changes requested):

ryansingman (approved):

PR #89, opened 2023-05-04
Fix join bugs with Pandas DataFrames

anishathalye:
df.join(other, on=...) always uses other's index, but uses the index column from df (an alternative is to set the index column for df as well, but that would mutate the df, which we want to preserve in this case). The rsuffix is not necessary: we don't expect there to be any other overlapping columns in cl_cols.

Later, when adding the __cleanlab_final_label column: we can't use cl_cols on the rhs of the assignment, because that DataFrame's index doesn't match the joined_ds, instead we should just use joined_ds.

Finally, we shouldn't mutate the DataFrame that was passed in, we should return a new DataFrame that has the corrections applied (later, we could expand the API to include an inplace=True kwarg).


ryansingman (approved):
Good catch, this LGTM!

PR #91, opened 2023-05-07
backwards compatible change to support better simple zip schema handling

axl1313:
### Description
This change is meant to support changes in https://github.com/cleanlab/cleanlab-studio-data-ingestion/pull/35 while being backwards compatible

### Motivation and Context
Detailed context here: https://www.notion.so/cleanlab/1271-d1d13ce18eeb4d0083001276c65a8a29?pvs=4

### How should this be tested?
- Check that uploading simple zip, zip w/ metadata, and text/tabular datasets work using cli and python API with just these changes
- Check that uploading simple zip, zip w/ metadata, and text/tabular datasets work using cli and python API with this change + data ingestion changes

### Checklist

- [x] PR was tested and change works as expected
- [ ] Design decisions and approaches were documented in the PR Description
- [ ] Design documents on Notion have been updated
- [ ] Tests (unit / end-to-end) are created for new utility functions or user flows

### DIFF
--- a/cleanlab_studio/cli/dataset/upload.py
+++ b/cleanlab_studio/cli/dataset/upload.py
@@ -64,7 +64,9 @@ def upload(
     else:
         schema = get_proposed_schema(api_key, upload_id)
         proceed_upload = None
-        if schema is not None:
+        if schema is None or schema.get("immutable", False):
+            proceed_upload = True
+        else:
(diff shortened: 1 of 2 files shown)


ryansingman (approved):
LGTM; need to bump version (after releasing ingestion changes)

PR #92, opened 2023-05-09
Feature: add support for creating projects through the python api

axl1313:
### Description
Adds support for creating projects through they Python API. User must supply dataset_id and project_name parameters. All other project fields are specified through optional kwargs. For fields that aren't supplied by the user, we make our best guess (similar to how the frontend supplies defaults)

### How to Test
- Test with backend changes in feat/python-api-project-creation
- Test that creating projects works for all modalities
- Try various combinations of user supplied and inferred project fields

### DIFF
--- a/cleanlab_studio/studio/studio.py
+++ b/cleanlab_studio/studio/studio.py
@@ -169,7 +169,7 @@ class Studio:
                     f"Invalid label column: {label_column}. Label column must have categorical feature type"
                 )
         else:
-            label_column = dataset_details["label_column_guess"]
+            label_column: str = dataset_details["label_column_guess"]
             print(f"Label column not supplied. Using best guess {label_column}")
 
         if feature_columns is not None and modality != "tabular":
@@ -178,8 +178,6 @@ class Studio:
             if modality == "tabular":
                 feature_columns = dataset_details["distinct_columns"]
                 print(f"Feature columns not supplied. Using all columns")
-            else:
-                feature_columns = []
 
         if text_column is not None:
             if modality != "text":
@@ -200,6 +198,6 @@ class Studio:
             modality=modality,
             modeltype=modeltype,
             label_column=label_column,
-            feature_columns=feature_columns,
+            feature_columns=feature_columns if feature_columns is not None else [],
             text_column=text_column,
         )
--- a/cleanlab_studio/studio/studio.py
+++ b/cleanlab_studio/studio/studio.py
@@ -150,9 +150,9 @@ class Studio:
         dataset_id: str,
         project_name: str,
         *,
-        tasktype: Optional[str] = "multi-class",
+        tasktype: str = "multi-class",
         modality: Optional[str] = None,
-        modeltype: Optional[str] = "regular",
+        modeltype: str = "regular",
         label_column: Optional[str] = None,
         feature_columns: Optional[List[str]] = None,
         text_column: Optional[str] = None,
@@ -169,7 +169,7 @@ class Studio:
                     f"Invalid label column: {label_column}. Label column must have categorical feature type"
                 )
         else:
-            label_column: str = dataset_details["label_column_guess"]
+            label_column = str(dataset_details["label_column_guess"])
             print(f"Label column not supplied. Using best guess {label_column}")
 
         if feature_columns is not None and modality != "tabular":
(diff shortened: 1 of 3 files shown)


cleanlab_studio/studio/studio.py
```diff
@@ -144,3 +144,60 @@ def apply_corrections(self, cleanset_id: str, dataset: Any) -> Any:
...
+        modeltype: str = "regular",
+        label_column: Optional[str] = None,
+        feature_columns: Optional[List[str]] = None,
+        text_column: Optional[str] = None,
+    ) -> str:
+        dataset_details = api.get_dataset_details(self._api_key, dataset_id)
+
+        if modality is None:
+            print("Modality not supplied. Using dataset modality")
+            modality = dataset_details["modality"]
```
ryansingman, line 164: thought: we want to deprecate dataset modality in the near future, and make modality a project-only construct. because of this, I don't think we want to make modality optional.

cleanlab_studio/studio/studio.py
```diff
@@ -144,3 +144,60 @@ def apply_corrections(self, cleanset_id: str, dataset: Any) -> Any:
...
+        self,
+        dataset_id: str,
+        project_name: str,
+        *,
+        tasktype: str = "multi-class",
+        modality: Optional[str] = None,
+        modeltype: str = "regular",
+        label_column: Optional[str] = None,
+        feature_columns: Optional[List[str]] = None,
+        text_column: Optional[str] = None,
```
ryansingman, line 158: since this is a fairly complex interface (with some non-obvious arguments) and is user-facing, we definitely need to add a docstring here. docstring should focus on the high-level meaning of each argument, from a user perspective

cleanlab_studio/studio/studio.py
```diff
@@ -144,3 +144,60 @@ def apply_corrections(self, cleanset_id: str, dataset: Any) -> Any:
...
+        else:
+            label_column = str(dataset_details["label_column_guess"])
+            print(f"Label column not supplied. Using best guess {label_column}")
+
+        if feature_columns is not None and modality != "tabular":
+            print("Project modality is not tabular. Ignoring feature columns")
+        if feature_columns is None:
+            if modality == "tabular":
+                feature_columns = dataset_details["distinct_columns"]
+                print(f"Feature columns not supplied. Using all columns")
```
ryansingman, line 180: nit: "Using all valid feature columns"

cleanlab_studio/studio/studio.py
```diff
@@ -144,3 +144,60 @@ def apply_corrections(self, cleanset_id: str, dataset: Any) -> Any:
...
+            if label_column not in dataset_details["label_columns"]:
+                raise ValueError(
+                    f"Invalid label column: {label_column}. Label column must have categorical feature type"
+                )
+        else:
+            label_column = str(dataset_details["label_column_guess"])
+            print(f"Label column not supplied. Using best guess {label_column}")
+
+        if feature_columns is not None and modality != "tabular":
+            print("Project modality is not tabular. Ignoring feature columns")
```
ryansingman, line 176: as a general principle, we should be failing on these types of things. if a user supplies an argument, they expect it to be used. if the argument can't be used, we should fail and alert the user. same principle applies for the text column conditional below

ryansingman (changes requested):
Mostly LGTM, some points to resolve before merging thouggh

cleanlab_studio/studio/studio.py
axl1313: Ah good call. That makes sense to me

cleanlab_studio/studio/studio.py
```diff
@@ -144,3 +144,70 @@ def apply_corrections(self, cleanset_id: str, dataset: Any) -> Any:
...
             return corrected_ds
+
+    def create_project(
+        self,
+        dataset_id: str,
+        project_name: str,
+        modality: str,
+        *,
+        tasktype: str = "multi-class",
+        modeltype: str = "regular",
```
ryansingman, line 155: suggestion: let's type these as literals instead of strings, to make clearer to the user what the choices are

ryansingman (approved):
One suggested improvement, otherwise LGTM. If you do make the change I suggested, no need for another review -- just go ahead and merge

PR #93, opened 2023-05-10
properly check client version in python api

axl1313:
### Description
Fix to properly check client version in python api when initializing Studio object. Also fixes issue with CLI introduced in #92 which led to a mismatch between the base urls used by the old cli `api_service` and new shared `api` module. Eventually, we should migrate the CLI to only use shared `api` module

### Testing
- tested by editing version locally to invalid version


ryansingman (commented):
suggestion: `check_client_version` sounds like it will raise an error if the check fails, not return a bool. would either rename (`is_valid_client_client_version`) or move the raise into the method

also, will this validation take effect for the CLI as well?

axl1313:
> suggestion: `check_client_version` sounds like it will raise an error if the check fails, not return a bool. would either rename (`is_valid_client_client_version`) or move the raise into the method
> 
> also, will this validation take effect for the CLI as well?

CLI already had correct validation haha

ryansingman (approved):

PR #94, opened 2023-05-10
Fix: make sure feature columns and label column don't overlap

axl1313:
### Description
Creating a tabular project w/ the python api and using the default for feature columns will fail right now since the full list of distinct columns from dataset details contains the label column. This leads to a duplicate column error during training. This PR removes the label column from `feature_columns` in the default case and adds validation to raise an error if user supplied feature columns overlap with label column


ryansingman (approved):
Good catch!

PR #96, opened 2023-05-11
Feature: cleanset status check and e2e demo

axl1313:
### Description
This PR adds a method to the Python API to get the cleanset ID for a project. It also adds a method to poll for cleanset status until training is complete. Additionally, it adds an end to end demo that goes through the steps of authenticating w/ an API key, uploading a dataset from a pandas dataframe, creating a project, getting the cleanset ID, and downloading cleanlab columns from the cleanset once training is complete

Note: adding tutorials to docs is not quite working yet. will revisit tomorrow morning

### Testing
Test w/ backend changes in `feat/python-api-cleanset-status` (deployed to dev)
- test using demo notebook
- can try other datasets as well


cleanlab_studio/examples/python_api_demo.ipynb
```diff
@@ -0,0 +1,309 @@
+{
```
ryansingman, line 1: since this is a tutorial / demo and not source code, we don't want to include this as part of the package. instead, we want to put this as part of our docs (see how the CLOS does it: https://github.com/cleanlab/cleanlab/tree/master/docs/source/tutorials)

cleanlab_studio/examples/python_api_demo.ipynb
```diff
@@ -0,0 +1,309 @@
...
+    "cleanlab_columns_df.head()"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": []
+  }
```
ryansingman, line 285: (not sure if i'm highlighting the right thing, but you have an empty cell at the bottom of the ipynb)

cleanlab_studio/internal/api/api.py
```diff
@@ -207,19 +208,19 @@ def clean_dataset(
     api_key: str,
     dataset_id: str,
     project_name: str,
-    tasktype: str,
+    task_type: str,
     modality: str,
-    modeltype: str,
+    model_type: str,
```
ryansingman, line 213: comment: I'm fine w this change for now, bc we have basically no users of this, but in the future we should make sure we have a strong reason to change user-facing APIs

cleanlab_studio/studio/clean.py
```diff
@@ -0,0 +1,33 @@
...
+                if status.startswith("FOLD"):
+                    pbar.update(int(FOLD_STATUS) - pbar.n)
+                else:
+                    pbar.update(int(status) - pbar.n)
+                for c in "|/-\\":
+                    time.sleep(0.1)
+                    pbar.set_description_str(f"Cleanset Progress: {c}")
+                status = api.get_project_status(api_key, project_id)
+            pbar.update(pbar.total - pbar.n)
+    return api.get_latest_cleanset_id(api_key, project_id)
```
ryansingman, line 33: Is there any reason to have this be one function? I think this code becomes much more clear if we split it up into discrete `poll` and `get` steps

cleanlab_studio/studio/studio.py
```diff
@@ -210,10 +210,18 @@ def create_project(
...
+            task_type=task_type,
             modality=modality,
-            modeltype=modeltype,
+            model_type=model_type,
             label_column=label_column,
             feature_columns=feature_columns if feature_columns is not None else [],
             text_column=text_column,
         )
+
+    def get_latest_cleanset_id(self, project_id: str) -> Optional[str]:
```
ryansingman, line 221: function name doesn't indicate that the function is blocking. I personally would _prefer_ splitting this up into `poll` and `get` steps (like I commented for the util functions above), but at the least, the function name needs to indicate that it's blocking.

an alternate approach here that's fairly idiomatic is to use a flag for `blocking`. if the user sets `blocking=True`, this function is blocking and waits until a cleanset ID is ready. if the user sets `blocking=False`, we return None if the cleanset ID isn't ready. would also recommend adding a timeout argument

ryansingman (changes requested):

cleanlab_studio/studio/studio.py
```diff
@@ -210,10 +214,22 @@ def create_project(
...
+            task_type=task_type,
             modality=modality,
-            modeltype=modeltype,
+            model_type=model_type,
             label_column=label_column,
             feature_columns=feature_columns if feature_columns is not None else [],
             text_column=text_column,
         )
+
+    def poll_cleanset_status(self, cleanset_id: str, timeout: Optional[int] = None) -> bool:
```
ryansingman, line 225: Even though it's already fairly clear, I would put in docstring that this operation will block until cleanset is ready, errored, or timeout has been reached.

docs/source/guide/python_api.rst
```diff
@@ -56,8 +56,43 @@ You can create a project for a dataset given the dataset ID returned from the up
     >>> studio.create_project(dataset_id, project_name, modality)
     <project ID>
 
+Get Cleanset ID
+---------------
+You can get the ID of the latest cleanset for your project given the project ID returned from the create project step.
```
ryansingman, line 61: Should state "why" the user is getting their cleanset ID (e.g it's needed to export results)

docs/source/quickstart/index.rst
```diff
@@ -127,6 +127,17 @@ Results can either be exported by:
 
 To export your results, you must first obtain the cleanset ID for your project.
 Your cleanset ID can be found by clicking `Export Cleanset` on your cleanset page then `Export Using API`.
+You can also use the use the Python API to get your cleanset ID:
```
ryansingman, line 130: Let's remove the above ways -- we want this to be an e2e Python flow

docs/source/index.rst
```diff
@@ -21,3 +22,4 @@ Cleanlab Studio automatically finds and fixes errors in your dataset to turn unr
 This package provides a Python API and CLI to interact with the Cleanlab Studio app.
 
 .. include:: /quickstart/index.rst
+.. include:: /tutorials/index.rst
```
ryansingman, line 25: I don't think we want to display this (this is just a `Tutorials` header w/ nothing underneath) on our index page. just adding to the toctree should be sufficient to have it appear in the sidebar

docs/source/tutorials/python_api_demo.ipynb
```diff
@@ -0,0 +1,330 @@
...
+    "3. Upload dataset to Studio"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stderr",
```
ryansingman, line 78: why is this showing up in stderr? (can see in docs that this is highlighted in red)

is this the case for the actual output? easy fix is to just change this to stdout

docs/source/tutorials/python_api_demo.ipynb
```diff
@@ -0,0 +1,330 @@
...
+    "6. Poll for cleanset status until cleanset is ready"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 7,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stderr",
```
ryansingman, line 174: as above wrt stderr

ryansingman (changes requested):

docs/source/tutorials/python_api_demo.ipynb
axl1313: not sure. might be a tqdm thing. i'll just edit to stdout

ryansingman: https://tqdm.github.io/docs/tqdm/#tqdmstd

yeah does seem to be tqdm

ryansingman (approved):

PR #99, opened 2023-05-15
add support for applying corrections to dataset w auto gen id

axl1313:
### Description
This PR adds support for running apply_corrections for a project with autogenerated ID columns for both pyspark and pandas dataframes. For projects created from datasets w/ autogenerated IDs we just join the corrected labels with the original dataframe in sorted order

### Motivation
We want to make autogenerated ID columns the default as we move away from requiring user specified ID columns

### Testing
For both pyspark and pandas datasets:
- Upload a dataset specifying id_column="cleanlab-autogenerate-id"
- Create a project and make some corrections. Also use exclude action for some rows
- Run `studio.apply_corrections()`. Check that rows w/ corrections have corrected label and rows w/ exclude action are properly excluded

### DIFF
--- a/cleanlab_studio/studio/studio.py
+++ b/cleanlab_studio/studio/studio.py
@@ -163,6 +163,11 @@ class Studio:
             corrected_ds = corrected_ds[joined_ds["action"] != "exclude"]
             return corrected_ds
 
+        else:
+            raise ValueError(
+                f"Provided unsupported dataset of type: {type(dataset)}. We currently support applying corrections to pandas or pyspark dataframes"
+            )
+
     def create_project(
         self,
         dataset_id: str,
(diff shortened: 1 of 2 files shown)


cleanlab_studio/studio/studio.py
```diff
@@ -133,16 +142,18 @@ def apply_corrections(self, cleanset_id: str, dataset: Any) -> Any:
...
-                dataset.drop(label_column)
+            return (
+                corrected_ds.drop(label_column)
                 .join(new_labels, on=id_col, how="right")
                 .where(new_labels["action"] != "exclude")
                 .drop("action")
             )
-            return corrected_df
-
         elif isinstance(dataset, pd.DataFrame):
```
ryansingman, line 151: nit: this is separate to the changes you've made here, but can you add a default case to this conditional for if the dataset is not a pyspark or pandas dataframe? currently, we'll just return None -- we should instead raise a value error. would be good if this func was typed correctly as well (dataset should not be `Any`)

ryansingman (approved):
LGTM overall! one nit

PR #100, opened 2023-05-24
adding outlier columns

clu0:
Let outlier column to be downloaded through cli.

Kat and I have tested this. @anishathalye can you double check the new documentation in the concepts page, and @axl1313 can you have a look at the code?


cleanlab_studio/studio/studio.py
```diff
@@ -69,17 +70,17 @@ def _download_cleanlab_columns(
     ) -> pd.DataFrame:
         rows = api.download_cleanlab_columns(self._api_key, cleanset_id, all=True)
         id_col = api.get_id_column(self._api_key, cleanset_id)
-        # TODO actually get _all_ the columns incl e.g., cleanlab_top_labels
-        # and have the API give the column headers rather than this library hard-coding it
```
anishathalye, line 73: "and have the API give the column headers rather than this library hard-coding it" - can we do that? The backend should just supply the cleanlab columns to the CLI, without duplicating the list of headers here, so when we add new columns, we don't need to change it in two places + release a new version of the CLI.

docs/source/concepts/cl_columns.rst
```diff
@@ -0,0 +1,38 @@
...
+Cleanlab Columns
+****************
+
+Introduction
+============
+
+Cleanlab Studio will run a number of analyses on your dataset to help identify potential data issues. Currently we look for the following issues in datasets:
+- Label issues
+- Outliers
+We will also continue to add new analyses soon. The raw outputs of each analysis will be available to download as Cleanlab columns by calling ``studio.download_cleanlab_columns()``.
```
anishathalye, line 12:
```suggestion
New analyses are being continuously added to Cleanlab Studio. The raw outputs of each analysis will be available to download as Cleanlab columns by calling ``studio.download_cleanlab_columns()``.
```

docs/source/concepts/cl_columns.rst
```diff
@@ -0,0 +1,38 @@
...
+
+For each analysis we will often expose two columns, one being a numeric score between 0 and 1 indicating the quality of a sample for that analysis (lower score means lower quality), and the other being a boolean, which indicates whether the sample is likely to have an issue or not. The different analyses are run separately, and hence they should be independently. For example, samples with label issues are not neccessarily outliers, and vice versa, and it is possible to have a sample with label issue marked as ``False``, but outlier marked as ``True``.
```
anishathalye, line 14:
```suggestion
Most analyses produce two columns, one being a numeric score between 0 and 1 indicating the quality of a sample for that analysis (lower score means lower quality), and the other being a boolean, which indicates whether the sample is likely to have an issue or not. The different analyses are run separately, and hence they should be treated independently. For example, samples with label issues are not necessarily outliers, and vice versa, and it is possible to have a sample with label issue marked as ``False``, but outlier marked as ``True``. Furthermore, it is possible to have samples with label issue marked as ``True` **and** outlier marked as ``True``: these are the raw results from the analyses, and you may interpret them as appropriate (e.g., if a data point is an outlier, it may be irrelevant that it's also likely a label issue).
```

docs/source/concepts/cl_columns.rst
```diff
@@ -0,0 +1,38 @@
...
+Label Issues
+============
+
+Label quality
+-------------
+The ``label_quality`` column contains a score bounded between 0 and 1. The score is calculated using confident learning. The lower the score of a sample, the more likely it has a label issue.
+
+Issue
+-----
+The ``issue`` column contains a boolean value, with ``True`` indicating that the sample is likely to have a label issue. The value is obtained by thresholding the ``label_quality`` score with confident learning.
```
anishathalye, line 27: Rename to `label_issue`? Just `issue` will be confusing as we add more analyses (already have > 1).

This is especially important to consider for the future, when we add more analyses: should we have a consistent scheme for naming Cleanlab columns? For the boolean, doing `{name of analysis}_issue` seems good, though `outlier_issue` sounds a bit weird (just `outlier` is better). For the continuous score, using the word "quality" or "score" seems good; choosing one is a bit tricky as well. "label_quality" sounds good, "label_score" doesn't sound quite as good; however, "outlier_quality" doesn't make sense, whereas "outlier_score" is fine.

docs/source/concepts/cl_columns.rst
```diff
@@ -0,0 +1,38 @@
...
+============
+
+Cleanlab Studio will run a number of analyses on your dataset to help identify potential data issues. Currently we look for the following issues in datasets:
+- Label issues
+- Outliers
+We will also continue to add new analyses soon. The raw outputs of each analysis will be available to download as Cleanlab columns by calling ``studio.download_cleanlab_columns()``.
+
+For each analysis we will often expose two columns, one being a numeric score between 0 and 1 indicating the quality of a sample for that analysis (lower score means lower quality), and the other being a boolean, which indicates whether the sample is likely to have an issue or not. The different analyses are run separately, and hence they should be independently. For example, samples with label issues are not neccessarily outliers, and vice versa, and it is possible to have a sample with label issue marked as ``False``, but outlier marked as ``True``.
+
+We give more details on the different Cleanlab columns below.
```
anishathalye, line 16: Presenting a summary in a rST [table](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/tables.html) would be really helpful, with details below (and perhaps hyperlinks from the column names to the details below).

docs/source/concepts/projects.rst
```diff
@@ -0,0 +1,27 @@
...
+
+Projects
+********
+
+Machine Learning Tasks
+======================
+Cleanlab Studio supports the following ML tasks:
+
+* Multi-class classification (:code:`multi-class`)
+* Multi-label classification (:code:`multi-label`) **ALERT: make sure this is released before the docs**
```
anishathalye, line 11: This "ALERT" line can be removed now.

anishathalye (commented):
Left some comments on the implementation and docs. If any points are easiest to discuss over Slack/Zoom, we can do that. Haven't tried running the code myself yet, but let me know if that would be helpful to do.

cleanlab_studio/cli/cleanset/download.py
```diff
@@ -62,15 +63,10 @@ def download(
     CleanlabSettings.init_cleanlab_dir()
     api_key = config.get_api_key()
     progress("Downloading Cleanlab columns...")
-    rows = api_service.download_cleanlab_columns(api_key, cleanset_id=id, all=all)
-
-    if all:
```
kat-wicks, line 67: It doesn't seem like this "all" argument is actually of use anymore -- consider getting rid of it?

kat-wicks (approved):

cleanlab_studio/studio/studio.py
clu0: sounds good, this is now resolved

docs/source/concepts/cl_columns.rst
clu0: renamed according to our slack discussion:
- add "is_{...}" for boolean
- add "{...}_score" for floats

docs/source/concepts/cl_columns.rst
clu0: added table

cleanlab_studio/cli/cleanset/download.py
clu0: Seems to have been removed due to the refactor to have names on backend side

docs/source/concepts/cl_columns.rst
```diff
@@ -0,0 +1,57 @@
...
+   * - :ref:`suggested_label <suggested_label>`
+     - String
+   * - :ref:`is_outlier <is_outlier>`
+     - Boolean
+
+Label Issues
+============
+
+.. _label_issue_score:
+``label_issue_score``
```
LukeMainwaring, line 37: should this be `label_quality_score` ? i'd expect a higher `label_issue_score` to be more likely to have a label issue

LukeMainwaring (approved):

LukeMainwaring:
no issues with the code changes, left one comment regarding the `label_issue_score` naming. I haven't tested these changes but can do so locally or once merged if some extra testing would be helpful

PR #101, opened 2023-05-26
Feature: add support for project deletion from python api

axl1313:
This PR adds a method to the Python API for deleting projects

### How to test
Test w/ backend branch `feat/python-api-project-deletion`
- Check that deleting one of your projects succeeds
- Check that attempting to delete project w invalid ID/project you don't own fails

### DIFF
--- a/cleanlab_studio/internal/api/api.py
+++ b/cleanlab_studio/internal/api/api.py
@@ -261,6 +261,11 @@ def get_cleanset_status(api_key: str, cleanset_id: str) -> JSONDict:
     return status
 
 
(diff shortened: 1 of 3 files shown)


ryansingman (approved):
Changes LGTM

quick question: do we want to document this in our RTD? or is this intended as a primarily internal functionality?

axl1313:
> Changes LGTM
> 
> quick question: do we want to document this in our RTD? or is this intended as a primarily internal functionality?

I think it can be primarily intended as internal functionality for now. The main use case is for deleting projects created while running automated benchmarking

PR #104, opened 2023-06-14
fix: pd read_json has error when id_col = 'index'; change to doing re…

clu0:
…name after read_json

related fix for https://github.com/cleanlab/cleanlab-studio-backend/pull/501, see that one for context


kat-wicks (approved):
lgtm cheers

PR #106, opened 2023-06-24
Fix/none vals

clu0:
Current handling of None values in `apply_corrections` wouldn't work if the none value is not a string with value "None". Added a more careful check for this.

### DIFF
--- a/cleanlab_studio/studio/studio.py
+++ b/cleanlab_studio/studio/studio.py
@@ -13,7 +13,7 @@ except ImportError:
 
 from . import clean, upload
 from cleanlab_studio.internal.api import api
-from cleanlab_studio.internal.util import init_dataset_source
+from cleanlab_studio.internal.util import init_dataset_source, check_none, check_not_none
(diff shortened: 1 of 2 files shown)


cleanlab_studio/studio/studio.py
```diff
@@ -118,12 +118,13 @@ def apply_corrections(self, cleanset_id: str, dataset: Any) -> Any:
             else:
                 joined_ds = dataset.join(cl_cols.set_index(id_col).sort_values(by=id_col))
             joined_ds["__cleanlab_final_label"] = joined_ds["clean_label"].where(
-                joined_ds["clean_label"] != "None", dataset[label_column]
+                np.asarray(list(map(check_not_none, joined_ds["clean_label"].to_numpy()))),
```
axl1313, line 121: nit: I believe you can just do `np.asarray(list(map(check_not_none, joined_ds["clean_label"])))`

axl1313 (approved):
code lgtm! did not test myself though (wasn't sure how to produce example where this new none check was necessary)

PR #108, opened 2023-06-27
improve null handling

clu0:
Switched the type casts from backend (`str` to `"string"`, `np.int64` to `"Int64"`), so we now need to handle NA for missing values


axl1313 (commented):
The pandas change lgtm! Do you have bandwidth to check if this affects the pyspark version as well? If not, I can take a look sometime

PR #111, opened 2023-07-06
Feat/telemetry

clu0:
decorator that wraps functions we want to track and sends stack trace to backend if error raised

backend pr: https://github.com/cleanlab/cleanlab-studio-backend/pull/544

### DIFF
--- a/cleanlab_studio/internal/api/api.py
+++ b/cleanlab_studio/internal/api/api.py
@@ -5,6 +5,7 @@ import functools
 import sys
 import traceback
 import json
+import contextlib
 
 from cleanlab_studio.errors import APIError
 
@@ -37,11 +38,12 @@ def telemetry(func: Callable[..., Any]) -> Callable[..., Any]:
             result = func(api_key, *args, **kwargs)
             return result
         except Exception as err:
-            _ = requests.post(
-                f"{cli_base_url}/telemetry",
-                data=traceback.format_exc(),
-                headers=_construct_headers(api_key),
-            )
+            with contextlib.suppress(Exception):
+                _ = requests.post(
+                    f"{cli_base_url}/telemetry",
+                    data=traceback.format_exc(),
+                    headers=_construct_headers(api_key),
+                )
             raise err
 
     return tracked_func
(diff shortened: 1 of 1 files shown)


cleanlab_studio/internal/api/api.py
```diff
@@ -18,6 +23,30 @@
...
+        print("using telemetry")
+        try:
+            result = func(api_key, *args, **kwargs)
+            return result
+        except Exception as err:
+            _ = requests.post(
+                f"{cli_base_url}/telemetry",
+                data=traceback.format_exc(),
+                headers=_construct_headers(api_key),
+            )
```
ryansingman, line 44: we don't want failed telemetry to ever result in an error visible to users, so let's make sure that this is wrapped in a try-except. one example I can think of is if the user is having network issues. their first request will error out, then they'll see a network error raised from a telemetry request. this would be a bit confusing to users -- best to avoid that

ryansingman (commented):
Mostly LGTM! See comment in this PR + backend comments

cleanlab_studio/internal/api/api.py
clu0: ah thanks that's a good point.

Fixed this, and also added function name and args to be sent to backend

clu0: if this looks good now, I'll add it to all the functions that make api calls to our backend

cleanlab_studio/internal/api/api.py
```diff
@@ -18,6 +24,39 @@
...
+    def tracked_func(api_key: str, *args: Any, **kwargs: Any) -> Any:
+        print("using telemetry")
+        try:
+            result = func(api_key, *args, **kwargs)
+            return result
+        except Exception as err:
+            with contextlib.suppress(Exception):
+                arg_list = [repr(arg) for arg in args]
+                arg_list.extend(f"{k}={v!r}" for k, v in kwargs.items())
+                arg_str = ", ".join(arg_list)
```
ryansingman, line 44: consider that this is considerably more invasive than just sending the traceback. would recommend checking w Anish to see if this is what we want to do at the moment

PR #113, opened 2023-07-11
convert directly to spark dataframe from download

clu0:
convert directly from json to pyspark dataframe in download, avoid going through pandas dataframe and dealing with <NA>

### DIFF
--- a/cleanlab_studio/internal/api/api.py
+++ b/cleanlab_studio/internal/api/api.py
@@ -1,6 +1,6 @@
 import os
 import time
-from typing import Any, Callable, List, Optional, Tuple, Dict
+from typing import Callable, List, Optional, Tuple, Union, Any
 from cleanlab_studio.errors import APIError
 
 import requests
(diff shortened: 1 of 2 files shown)


cleanlab_studio/internal/api/api.py
```diff
@@ -155,25 +162,44 @@ def get_label_column_of_project(api_key: str, project_id: str) -> str:
...
 
     :param api_key:
     :param cleanset_id:
     :param all: whether to download all Cleanlab columns or just the clean_label column
-    :return: return (rows, id_column)
+    :return: return a dataframe, either pandas or spark. Type is Any because don't want to require spark installed
     """
     res = requests.get(
         cli_base_url + f"/cleansets/{cleanset_id}/columns?all={all}",
+        params=dict(to_spark=to_spark),
```
axl1313, line 181: nit: can we move the `all` query param to this argument as well instead of adding it to the url manually

axl1313 (commented):
code lgtm! just wanted to confirm that you tested that the pandas version still works as expected and that the pyspark version works in databricks? will approve on confirmation

(also have one small nit)

clu0:
> code lgtm! just wanted to confirm that you tested that the pandas version still works as expected and that the pyspark version works in databricks? will approve on confirmation
> 
> (also have one small nit)

yup things still work

cleanlab_studio/internal/api/api.py
clu0: sounds good updated

axl1313 (approved):
lgtm!

PR #117, opened 2023-07-21
Feature/inference api

taekang1618:
From one of our client calls, we found a demand to use model deployment in automatic data pipelines using the studio python api client. In this PR, we built the api client side code for model deployment.

To test, you need to need to pull the branch in this [PR](https://github.com/cleanlab/cleanlab-studio-backend/pull/588) from the backend code.


ryansingman:
can you fix mypy checks?

ryansingman (changes requested):
this PR is incomplete. users are expected to use the `Studio` class to interact w/ the API, not the API methods directly. methods need to be added to that class that call the API methods (like all of the other functionalities do)

cleanlab_studio/internal/api/api.py
```diff
@@ -329,3 +333,64 @@ def poll_progress(
...
+    )
+
+    handle_api_error(res)
+    presigned_url = res.json()["upload_url"]
+    query_id: str = res.json()["query_id"]
+    modality = res.json()["modality"]
+    header = res.json()["header"]
+    if modality == "text":
+        header_io = io.StringIO(header)
+        input_batch = io.StringIO("\n".join(chain(header_io, batch)))
```
ryansingman, line 353: see comment in backend

cleanlab_studio/internal/api/api.py
```diff
@@ -329,3 +333,64 @@ def poll_progress(
...
+    elif status == "FAILED":
+        return {"status": "error", "error_msg": error_msg}
+    else:
+        return {"status": "running"}
+
+
+def download_prediction_results(result_url: str) -> io.StringIO:
+    """Downloads prediction results from presigned URL."""
+    res = requests.get(result_url)
+    return io.StringIO(res.text)
```
ryansingman, line 396: no need to do this buffering, just pass `res.raw`

cleanlab_studio/studio/inference.py
```diff
@@ -0,0 +1,90 @@
...
+        :param batch: batch of example to predict classes for, as in-memory CSV file
+        :return: predictions from batch
+        """
+        query_id: str = api.upload_predict_batch(self._api_key, self._model_id, batch)
+        api.start_prediction(self._api_key, query_id)
+
+        resp = api.get_prediction_status(self._api_key, query_id)
+        status: Optional[str] = resp["status"]
+        # Set timeout to 10 minutes as inference won't take longer than 10 minutes typically and
+        # to prevent users from getting stuck in this loop indefinitely when there is a failure
```
ryansingman, line 55: failures should be reported in the status, correct?

cleanlab_studio/studio/inference.py
```diff
@@ -0,0 +1,90 @@
...
+        timeout = time.time() + 60 * 10
+        while status == "running" and time.time() < timeout:
+            resp = api.get_prediction_status(self._api_key, query_id)
+            status = resp["status"]
+
+        if status == "error":
+            return resp["error_msg"]
```
ryansingman, line 62: we need to raise an error here. returning an error message is not a sensible API for users

cleanlab_studio/studio/inference.py
```diff
@@ -0,0 +1,90 @@
...
+        else:
+            result_url = resp["result_url"]
+            results: io.StringIO = api.download_prediction_results(result_url)
+            results_converted: Predictions = pd.read_csv(results).to_numpy()
+            return results_converted
+
+    @functools.singledispatchmethod
```
ryansingman, line 69: should remove (relic from my dev work)

cleanlab_studio/studio/inference.py
```diff
@@ -0,0 +1,90 @@
...
+        :return: predictions from batch
+        """
+        query_id: str = api.upload_predict_batch(self._api_key, self._model_id, batch)
+        api.start_prediction(self._api_key, query_id)
+
+        resp = api.get_prediction_status(self._api_key, query_id)
+        status: Optional[str] = resp["status"]
+        # Set timeout to 10 minutes as inference won't take longer than 10 minutes typically and
+        # to prevent users from getting stuck in this loop indefinitely when there is a failure
+        timeout = time.time() + 60 * 10
```
ryansingman, line 56: timeout should be user configurable

ryansingman (changes requested):

cleanlab_studio/internal/api/api.py
taekang1618: Got it. I'll change the lambda function to take care of this. Didn't want to touch lambda as much as possible since it's not in a state that's easy to fix and deploy right now.

cleanlab_studio/internal/api/api.py
taekang1618: I think we do need this buffering since we want to return a numpy array for the results to the user.

cleanlab_studio/studio/inference.py
taekang1618: Yeah that would be ideal, but there has been cases where the status updates did not happen properly for lambda failures in the past and queries were stuck in the `running` status even after they have failed. Just wanted to make sure this case doesn't confuse users.

cleanlab_studio/studio/inference.py
taekang1618: Changed to raising an APIError

cleanlab_studio/studio/inference.py
taekang1618: removed

cleanlab_studio/studio/inference.py
taekang1618: I'll add timeout as an optional parameter users can add when calling predict.

cleanlab_studio/internal/api/api.py
```diff
@@ -329,3 +333,57 @@ def poll_progress(
...
```
ryansingman, line 389: It is unnecessary and wasteful to rebuffer the entire result here. Just use res.raw

ryansingman (changes requested):

cleanlab_studio/internal/api/api.py
taekang1618: Removed this function entirely and used url directly in pandas to read results

ryansingman:
Failing in dev currently:
```
cleanlab_studio.errors.APIError: Length of values (2) does not match length of index (100)
```

taekang1618:
> 

Can you let me know which model and test input you tested with?

taekang1618:
Looked at the input file [here](https://s3.console.aws.amazon.com/s3/object/cleanlab-dev-base-stack-cleanlabstudiomodeldeploy-14ebhlpu4a08n?region=us-west-1&prefix=5db799c1981b4f3296bf5b1ad258be8a/input/cli_e3a7ba23-1bd9-4795-9821-a20900d86291) and it seems like there is a formatting error in the header. Also, I thought we weren;t allowing users to put headers in the input files. I'll include that in the python docstring as well so that users will be warned ahead of time.

cleanlab_studio/studio/studio.py
```diff
@@ -231,6 +231,10 @@ def delete_project(self, project_id: str) -> None:
         api.delete_project(self._api_key, project_id)
         print(f"Successfully deleted project: {project_id}")
 
+    def get_model(self, model_id: str) -> inference.Model:
+        """Creates model object from model ID, to use for inference."""
```
axl1313, line 235: Can you merge the latest `main` into this branch and update docstrings to match the format used there (there's a link to the docstring style guide in the README)? Also make sure to document all args and the return value for user facing functions

cleanlab_studio/studio/inference.py
```diff
@@ -0,0 +1,101 @@
...
+
+import numpy as np
+import numpy.typing as npt
+import pandas as pd
+
+from cleanlab_studio.errors import APIError
+from cleanlab_studio.internal.api import api
+
+
+TextBatch = Union[List[str], npt.NDArray[np.str_], pd.Series]
```
axl1313, line 16: You should be able to just use `TextBatch = Iterable[str]` here. TextBatch just needs to be some collection of strings you can iterate through right?

taekang1618: True.. although I think Ryan and Anish decided on the user interface, so I'm not sure if we want to support any iterable here.

taekang1618: https://www.notion.so/cleanlab/Model-Inference-Python-API-d5209e022f744298835010dbbfb26a4c

cleanlab_studio/studio/inference.py
```diff
@@ -0,0 +1,101 @@
...
+        resp = api.get_prediction_status(self._api_key, query_id)
+        status: Optional[str] = resp["status"]
+        # Set timeout to prevent users from getting stuck indefinitely when there is a failure
+        timeout_limit = time.time() + timeout
+
+        while status == "running" and time.time() < timeout_limit:
+            resp = api.get_prediction_status(self._api_key, query_id)
+            status = resp["status"]
+            # Set time.sleep so that the while loop doesn't flood backend with api calls
+            time.sleep(3)
```
ryansingman, line 68: make timeout configurable and cleanup logic

cleanlab_studio/internal/api/api.py
```diff
@@ -330,3 +334,51 @@ def poll_progress(
...
+        headers=_construct_headers(api_key),
+    )
+    handle_api_error(res)
+
+    prediction_results = res.json()
+    status = prediction_results["status"]
+    result_url = prediction_results["results"]
+    error_msg = prediction_results["error_msg"]
+
+    if status == "COMPLETE":
```
axl1313, line 379: Can we just use the status string returned by the api instead of redefining another set of hard coded statuses

cleanlab_studio/studio/studio.py
```diff
@@ -290,6 +290,18 @@ def delete_project(self, project_id: str) -> None:
...
 
+    def get_model(self, model_id: str) -> inference.Model:
+        """
+        Gets a model deployed by Cleanlab Studio.
+
+        Args:
+            model_id: ID of model to get. This ID should be fetched in the deployments page of the app UI.
+
+        Returns:
+            Model object with methods run predictions on new input data
```
axl1313, line 301: nit: methods *to* run predictions

axl1313 (approved):
lgtm! just a couple nits

ryansingman (approved):

PR #118, opened 2023-07-25
sort by cleanlab-generated ID if one exists

kat-wicks:
@sanjanag and @anishathalye ran into a problem where rows are returned in an arbitrary order, rather than being sorted by cleanlab ID  - see [https://github.com/cleanlab/cleanlab-studio-integration/issues/1113](https://github.com/cleanlab/cleanlab-studio-integration/issues/1113 ) for additional context.

This checks for our autogenerated ID column and returns it ordered by that column if one exists. 

"Design" notes: I didn't add defensive checks such as checking if integer etc. because cleanlab_row_ID is an internal column that users' shouldn't be messing with. I also didn't reindex because 1) unneccesary computation; could slow export for big datasets and 2) autogenerate ID is autoselected and many users have other ID columns :)

### DIFF
--- a/cleanlab_studio/studio/studio.py
+++ b/cleanlab_studio/studio/studio.py
@@ -72,8 +72,6 @@ class Studio:
                 rows_df = rows_df.drop("action")
             else:
                 rows_df.drop("action", inplace=True, axis=1)
-        if "cleanlab_row_ID" in rows_df.columns:
-            rows_df.sort_values(by="cleanlab_row_ID", inplace=True)
         return rows_df
 
     def apply_corrections(self, cleanset_id: str, dataset: Any, keep_excluded: bool = False) -> Any:
--- a/cleanlab_studio/studio/studio.py
+++ b/cleanlab_studio/studio/studio.py
@@ -72,6 +72,8 @@ class Studio:
                 rows_df = rows_df.drop("action")
             else:
                 rows_df.drop("action", inplace=True, axis=1)
+        if "cleanlab_row_ID" in rows_df.columns:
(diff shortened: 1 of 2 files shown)


cleanlab_studio/studio/studio.py
```diff
@@ -72,6 +72,8 @@ def download_cleanlab_columns(
                 rows_df = rows_df.drop("action")
             else:
                 rows_df.drop("action", inplace=True, axis=1)
+        if "cleanlab_row_ID" in rows_df.columns:
+            rows_df.sort_values(by="cleanlab_row_ID")
```
clu0, line 76: don't you need a `inplace=True` here?

kat-wicks: How odd. You're right and yet my testing worked. Must have been a fluke I suppose?

clu0 (approved):

anishathalye:
Would it make sense to sort by the ID column regardless (even if it's a custom ID column)? Could do that as part of this PR. See related Slack thread: https://cleanlabinc.slack.com/archives/C02SN6TFSF8/p1690644910271439?thread_ts=1690564231.467479&cid=C02SN6TFSF8

clu0:
> Would it make sense to sort by the ID column regardless (even if it's a custom ID column)? Could do that as part of this PR. See related Slack thread: https://cleanlabinc.slack.com/archives/C02SN6TFSF8/p1690644910271439?thread_ts=1690564231.467479&cid=C02SN6TFSF8

Yup this should be fixed with the new commit. @kat-wicks take a look and then we can merge?

PR #121, opened 2023-08-01
Fix/web cli export

kat-wicks:
Based on conversations with @ulya-tkch  and @anishathalye, we are updating the CLI, Python library, and Web export to have the same columns to reduce confusion.

Instead of somewhat strange "only ID + corrected label"/"all cleanlab columns"/"all cleanlab columns except action" export options, now have "only ID + corrected"/"all cleanlab columns"/"all project-related columns" for CLI/python export, which can be mixed and matched. This gives more information to CLI users , who previously were missing project information like actions taken, if a datapoint was initially unlabeled, etc. the "--all" flag is preserved.


clu0 (changes requested):
Code looks good to me, but if we are exporting some new columns (like "initially unlabeled", from looking at the backend code), we should add documentation for them in the concepts section

kat-wicks:
** test with Pyspark

docs/source/concepts/cl_columns.rst
```diff
@@ -124,3 +124,93 @@ Contains a boolean value, with ``True`` indicating that the sample is likely to
...
+
+
++-------------------------------+------------------------------------------------------------------+
+| Project Detail                | Cleanlab Column Name                              | Value Type   |
++===============================+==================================================================+
+| Action                        | :ref:`action <action>`                            | String       |
++-------------------------------+------------------------------------------------------------------+
+| Corrected Label               | :ref:`corrected_label <corrected_label>`          | String       |
++-------------------------------+------------------------------------------------------------------+
+| Previously Resolved           | :ref:`previously_resolved <previously_resolved>`  | Boolean      |
```
clu0, line 151: let's change all the names for booleans to `is_...`

docs/source/concepts/cl_columns.rst
```diff
@@ -124,3 +124,93 @@ Contains a boolean value, with ``True`` indicating that the sample is likely to
...
++-------------------------------+------------------------------------------------------------------+
+| Given Label                   | :ref:`given_label <given_label>`                  | String       |
++-------------------------------+------------------------------------------------------------------+
+| Likelihood of Given Label     | :ref:`p_given_label <p_given_label>`              | Float        |
++-------------------------------+------------------------------------------------------------------+
+| Predicted Label               | :ref:`predicted_label <predicted_label>`          | String       |
++-------------------------------+------------------------------------------------------------------+
+| Likelihood of Predicted Label | :ref:`p_predicted_label <p_predicted_label>`      | Float        |
++-------------------------------+------------------------------------------------------------------+
+| Initially Unlabeled           | :ref:`initially_unlabeled <initially_unlabeled>`  | Boolean      |
```
clu0, line 165: change to `is_`

docs/source/concepts/cl_columns.rst
```diff
@@ -124,3 +124,93 @@ Contains a boolean value, with ``True`` indicating that the sample is likely to
 -----------
 Contains a score bounded between 0 and 1, which is used to determine whether a sample is an outlier. The lower the score of a sample, the more likely it is to be an outlier.
 
+
+Project Information Columns
```
clu0, line 128: I don't think it makes sense to have the `Project Information Columns` be a subsection under the Cleanlab Columns page.

I think better to add a new page Project Information Columns, and link it directly under the "Concepts" page

docs/source/concepts/cl_columns.rst
```diff
@@ -124,3 +124,93 @@ Contains a boolean value, with ``True`` indicating that the sample is likely to
...
+============
+
+
+You can also retrieve information/metadata related to the Cleanlab project itself.
```
clu0, line 132: I think this line would be better if it were more descriptive. Maybe something like this?

For each project that you run in Cleanlab Studio, Cleanlab stores a number of metadata columns that can be retrieved using the Python API. Currently we store the following project information columns:

docs/source/concepts/project_information.rst
```diff
@@ -0,0 +1,95 @@
...
+Currently we store the following project information columns:
+
+- Action Taken
+- If a row previously had an issue
+- Top Labels and Probabilities (alternative suggested labels and how likely they are)
+- The originally assigned (given) label and how likely it is to be the true label
+- The predicted label and its likelihood (even when we don't suggest changing from the given label).
+- If the row was initially unlabeled/had no given label. 
+
+To get this information, use ``studio.download_cleanlab_columns()`` and add the argument -p or -a for project details or all columns including analysis columns respectively.
```
clu0, line 20: wait I don't understand what this line means. What is the argument you're supposed to pass into the `download_cleanlab_columns` function when you use the python CLI to get the project info columns?

clu0 (changes requested):
tests against dev still buggy

clu0 (approved):
seems to work now, lgtm

PR #125, opened 2023-08-18
Feat/inference arbitrary batches

kat-wicks:
Split each batch into sub-batches so that users don't have to deal with timeout problems when using inference API. 

See this notion doc: https://www.notion.so/cleanlab/Model-Inference-Python-API-d5209e022f744298835010dbbfb26a4c 

10,000 for text and 100,000 seems like a safe number based on my own testing, but I would appreciate help testing further.

### DIFF
--- a/cleanlab_studio/studio/inference.py
+++ b/cleanlab_studio/studio/inference.py
@@ -11,7 +11,7 @@ import pandas as pd
 
 from cleanlab_studio.internal.api import api
 
-BATCH_MAX_TEXT, BATCH_MAX_TABULAR = 100000, 100000
+BATCH_MAX_TEXT, BATCH_MAX_TABULAR = 10000, 100000
 TextBatch = Union[List[str], npt.NDArray[np.str_], pd.Series]
 TabularBatch: TypeAlias = pd.DataFrame
 Batch = Union[TextBatch, TabularBatch]
--- a/cleanlab_studio/studio/inference.py
+++ b/cleanlab_studio/studio/inference.py
@@ -4,6 +4,7 @@ import io
 import time
 from typing import List, Tuple, Union
 from typing_extensions import TypeAlias
+import logging
 
 import numpy as np
 import numpy.typing as npt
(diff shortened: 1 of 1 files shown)


cleanlab_studio/studio/inference.py
```diff
@@ -45,13 +48,27 @@ def predict(
         Returns:
             predictions from batch as a numpy array, optionally also pandas dataframe of class probabilties
         """
-        csv_batch = self._convert_batch_to_csv(batch)
-        predictions, class_probabilities = self._predict_from_csv(csv_batch, timeout)
+
+        batchsize: int = BATCH_MAX_TABULAR if isinstance(batch, pd.DataFrame) else BATCH_MAX_TEXT
+        batched_preds: List[Predictions] = []
+        batched_probs: List[ClassProbablities] = []
+        num_batches = len(batch) // batchsize + (1 if len(batch) % batchsize != 0 else 0)
```
LukeMainwaring, line 55: nit: you could use `math.ceil` here
`num_batches = math.ceil(len(batch) / batchsize)`

cleanlab_studio/studio/inference.py
```diff
@@ -45,13 +48,27 @@ def predict(
...
+        for i in range(num_batches):
+            start = i * batchsize
+            csv_batch = self._convert_batch_to_csv(
+                batch[start : min(start + batchsize, len(batch))]
+            )
+            predictions, class_probabilities = self._predict_from_csv(csv_batch, timeout)
+            batched_preds.append(predictions)
+            batched_probs.append(class_probabilities)
+        preds_type = int if isinstance(batched_preds[0][0], int) else str
```
LukeMainwaring, line 64: nit: this might be clearer if written as `preds_type = type(batched_preds[0][0])` but not too important

kat-wicks: I think that using `type` might be hinky because type() doesn't respect inheritance but isinstance() does and we have like pd.Int64 and idk different numpy strings that give me the heebie jeebies

LukeMainwaring: okay i see, that makes sense. let's keep it as-is right now

LukeMainwaring (approved):
lgtm!

LukeMainwaring (approved):
lgtm but make sure CI passes before merging. it looks like there's a typing issue somewhere

PR #128, opened 2023-09-12
Feature: wait until cleanset ready

ryansingman:
- adds wait_until_cleanset_ready method
- add deprecation warning to poll_cleanset_status


cleanlab_studio/errors.py
```diff
@@ -24,3 +24,7 @@ def __init__(self) -> None:
 class AuthError(APIError):
     def __init__(self) -> None:
         super().__init__("invalid API key")
+
+
+class CleansetRunError(RuntimeError):
```
anishathalye, line 29: Do we want a special exception for this case, for this single function? Maybe a more general exception that indicates that something went wrong, and that it's a permanent failure. Could be helpful, e.g., for a `deploy_model()` method where something goes wrong during training.

Could do, e.g., `class InternalError(Exception)`

cleanlab_studio/studio/studio.py
```diff
@@ -266,7 +268,30 @@ def poll_cleanset_status(self, cleanset_id: str, timeout: Optional[int] = None)
...
+
+        try:
+            clean_helpers.poll_cleanset_status(self._api_key, cleanset_id, timeout)
+            return True
+
+        except (TimeoutError, CleansetRunError):
+            return False
+
+    def wait_until_cleanset_ready(self, cleanset_id: str, timeout: Optional[float] = None) -> None:
+        """Blocks until cleanset is ready, timeout is reached, or cleanset errors.
```
anishathalye, line 284:
```suggestion
        """Blocks until a cleanset is ready or the timeout is reached.
```

anishathalye: I think the error case should be highlighted in the first line of the description. If our server code is bug-free, this should never happen (it's never a user error, always an internal server error).

cleanlab_studio/studio/studio.py
```diff
@@ -266,7 +268,30 @@ def poll_cleanset_status(self, cleanset_id: str, timeout: Optional[int] = None)
...
+
+        Args:
+            cleanset_id (str): ID of cleanset to check status or
+            timeout (Optional[float], optional): timeout for polling, in seconds. Defaults to None.
```
anishathalye, line 288: I don't remember what syntax we're following here, but why do we have "Optional[float]" as well as "optional"?

cleanlab_studio/studio/studio.py
```diff
@@ -266,7 +268,30 @@ def poll_cleanset_status(self, cleanset_id: str, timeout: Optional[int] = None)
...
+            return True
+
+        except (TimeoutError, CleansetRunError):
+            return False
+
+    def wait_until_cleanset_ready(self, cleanset_id: str, timeout: Optional[float] = None) -> None:
+        """Blocks until cleanset is ready, timeout is reached, or cleanset errors.
+
+        Args:
+            cleanset_id (str): ID of cleanset to check status or
```
anishathalye, line 287:
```suggestion
            cleanset_id (str): ID of cleanset to check status for.
```

cleanlab_studio/studio/studio.py
```diff
@@ -266,7 +268,30 @@ def poll_cleanset_status(self, cleanset_id: str, timeout: Optional[int] = None)
...
+            timeout (Optional[float], optional): timeout for polling, in seconds. Defaults to None.
+
+        Raises:
+            TimeoutError: if cleanset is not ready by end of timeout
+            CleansetRunError: if cleanset errored while running
```
anishathalye, line 292: Remember to update if renaming the exception

cleanlab_studio/errors.py
ryansingman: I expect we might want to use this error in the future in other places -- for example, a `train` method that takes in a pandas dataframe, uploads it, creates project, waits for and returns results.

I'll create an `InternalError` base class though and have this inherit from it

ryansingman: I expect we might want to use this error in the future in other places -- for example, a `train` method that takes in a pandas dataframe, uploads it, creates project, waits for and returns results.

I'll create an `InternalError` base class though and have this inherit from it

anishathalye: Ok. For this one, I'd suggest renaming to `CleansetError`

cleanlab_studio/studio/studio.py
ryansingman: `Optional` (uppercase `O`) means the value can be `None`.

`optional` (lowercase `o`) means the caller of this function doesn't have to supply a value here (as it defaults to None)

cleanlab_studio/errors.py
```diff
@@ -24,3 +24,11 @@ def __init__(self) -> None:
 class AuthError(APIError):
     def __init__(self) -> None:
         super().__init__("invalid API key")
+
+
+class InternalError(RuntimeError):
```
anishathalye, line 29: I think it's more common to inherit from `Exception`; I think `RuntimeError` is Python's "not any of the other built-in errors, like `ValueError` or `IndexError` or ...", but I usually see custom exceptions inheriting from either one of the specific Python errors (like `ValueError`) or directly from `Exception`.

cleanlab_studio/internal/clean_helpers.py
```diff
@@ -1,15 +1,17 @@
...
 from cleanlab_studio.internal.api import api
 
 
-def poll_cleanset_status(api_key: str, cleanset_id: str, timeout: Optional[int] = None) -> bool:
+def poll_cleanset_status(api_key: str, cleanset_id: str, timeout: Optional[float] = None) -> None:
     start_time = time.time()
     res = api.get_cleanset_status(api_key, cleanset_id)
-    spinner = "|/-\\"
-    spinner_idx = 0
+    spinner = itertools.chain("|/-\\")
```
anishathalye, line 14: Would be nice to factor out spinner into a reusable utility, rather than having special spinner code written or copy-pasted into these kinds of helper functions. But this could be done separately from this PR.

cleanlab_studio/internal/clean_helpers.py
```diff
@@ -19,22 +21,22 @@ def poll_cleanset_status(api_key: str, cleanset_id: str, timeout: Optional[int]
...
-            return True
+            return
+
         if res["has_error"]:
             pbar.set_postfix_str(res["step_description"])
-            return False
+            raise CleansetError(f"Cleanset {cleanset_id} failed to complete")
 
-    return False
+    return
```
anishathalye, line 42: Omit unnecessary return

cleanlab_studio/internal/clean_helpers.py
ryansingman: I started doing that, but didn't want to spend the extra time. This is the only place we use the spinner -- can refactor out the next time

cleanlab_studio/errors.py
```diff
@@ -24,3 +24,11 @@ def __init__(self) -> None:
 class AuthError(APIError):
     def __init__(self) -> None:
         super().__init__("invalid API key")
+
+
+class InternalError(BaseException):
```
anishathalye, line 29:
```suggestion
class InternalError(Exception):
```

https://docs.python.org/3/library/exceptions.html#module-exceptions

> The built-in exception classes can be subclassed to define new exceptions; programmers are encouraged to derive new exceptions from the [Exception](https://docs.python.org/3/library/exceptions.html#Exception) class or one of its subclasses, and not from [BaseException](https://docs.python.org/3/library/exceptions.html#BaseException). More information on defining exceptions is available in the Python Tutorial under [User-defined Exceptions](https://docs.python.org/3/tutorial/errors.html#tut-userexceptions).

anishathalye (approved):

PR #132, opened 2023-09-21
Fix bug with using chain instead of cycle

anishathalye:
`itertools.chain` is the wrong method to use, it creates a finite-length iterable, and so the code to display the spinner raises a `StopIteration` after going through the characters once.


anishathalye:
This needs to be merged before we make our next release, otherwise the broken https://github.com/cleanlab/cleanlab-studio/pull/128 will get released.

PR #139, opened 2023-10-19
Fix: fail when invalid api key is passed to Studio

ryansingman:
Previously, we called the `validate` endpoint but didn't check the result. This makes it such that there will be an immediate failure when a bad API key is passed, rather than waiting until another endpoint is called (and returning a bad error message)

Closes https://github.com/cleanlab/cleanlab-studio-integration/issues/1381

### DIFF
--- a/cleanlab_studio/studio/studio.py
+++ b/cleanlab_studio/studio/studio.py
@@ -1,6 +1,7 @@
 """
 Python API for Cleanlab Studio.
 """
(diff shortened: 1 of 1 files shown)


ryansingman:
Closes: https://github.com/cleanlab/cleanlab-studio-integration/issues/1381

axl1313 (approved):

PR #141, opened 2023-10-28
Changes to `apply_corrections` in studio.py

AnimatorJoe:
Addressing https://github.com/cleanlab/cleanlab-studio-integration/issues/1388

- fixed current failed caused because "clean_label" column has been renamed to "corrected_label"
- cleaned up [null check](https://github.com/cleanlab/cleanlab-studio/blob/017c5e37c3b2a5094dbe3e60303b0a19679d3344/cleanlab_studio/studio/studio.py#L176)

### DIFF
--- a/cleanlab_studio/studio/studio.py
+++ b/cleanlab_studio/studio/studio.py
@@ -95,7 +95,8 @@ class Studio:
 
         Args:
             cleanset_id: ID of cleanset to download columns from. To obtain cleanset ID from project ID use, [get_latest_cleanset_id](#method-get_latest_cleanset_id).
-            include_project_details: Whether to include a column with any actions taken on the cleanset in the downloaded columns.
+            include_cleanlab_columns: whether to download all Cleanlab columns or just the clean_label column
+            include_project_details: whether to download columns related to project status such as resolved rows, actions taken, etc.
 
         Returns:
             A pandas or pyspark DataFrame. Type is `Any` to avoid requiring pyspark installation.
@@ -173,7 +174,7 @@ class Studio:
             else:
                 joined_ds = dataset.join(cl_cols.set_index(id_col).sort_values(by=id_col))
             joined_ds["__cleanlab_final_label"] = joined_ds["corrected_label"].where(
-                joined_ds["corrected_label"].apply(check_not_none).tolist(),
+                joined_ds["corrected_label"].notnull().tolist(),
                 dataset[label_column].tolist(),
             )
 
--- a/cleanlab_studio/studio/studio.py
+++ b/cleanlab_studio/studio/studio.py
@@ -142,7 +142,7 @@ class Studio:
                     id_col,
                     row_number().over(Window.orderBy(monotonically_increasing_id())) - 1,
                 )
-            both = cl_cols.select([id_col, "action", "clean_label"]).join(
+            both = cl_cols.select([id_col, "action", "corrected_label"]).join(
                 corrected_ds_spark.select([id_col, label_column]),
                 on=id_col,
                 how="left",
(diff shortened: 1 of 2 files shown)


cleanlab_studio/studio/studio.py
```diff
@@ -166,15 +166,15 @@ def apply_corrections(self, cleanset_id: str, dataset: Any, keep_excluded: bool
...
             if id_col in dataset.columns:
                 joined_ds = dataset.join(cl_cols.set_index(id_col), on=id_col)
             else:
                 joined_ds = dataset.join(cl_cols.set_index(id_col).sort_values(by=id_col))
-            joined_ds["__cleanlab_final_label"] = joined_ds["clean_label"].where(
-                np.asarray(list(map(check_not_none, joined_ds["clean_label"].to_numpy()))),
-                dataset[label_column].to_numpy(),
+            joined_ds["__cleanlab_final_label"] = joined_ds["corrected_label"].where(
+                joined_ds["corrected_label"].apply(check_not_none).tolist(),
+                dataset[label_column].tolist(),
```
axl1313, line 177: Can you try to check if we need to do `.apply(check_not_none)`? We might be able to just use pandas `.isna()` since we [convert](https://github.com/cleanlab/cleanlab-studio-backend/blob/27ddbf189c8cad66def1fa371caf2a34c89dd997/backend/utils/export/export_util.py#L204) columns to dtypes where null values are represented by [`pd.NA`](https://pandas.pydata.org/docs/user_guide/missing_data.html#experimental-na-scalar-to-denote-missing-values) in the backend. Since we use json to encode the result in the api response, we shouldn't lose any information about null values. 

Let me know if you need any guidance on how to verify this. Would be good to check for both pyspark and pandas cases (I can get you a databricks account so you can test the pyspark version).

jwmueller:
@AnimatorJoe Great to see your first contribution. 
Please do sign the CLA linked above (only need to do it once), thanks!

cleanlab_studio/studio/studio.py
```diff
@@ -153,7 +154,7 @@ def apply_corrections(self, cleanset_id: str, dataset: Any, keep_excluded: bool
                 # instead, use original JSON, which uses null values where it's not specified
                 udf(lambda original, clean: original if check_none(clean) else clean)(
                     both[label_column],
-                    "clean_label",
+                    "corrected_label",
```
axl1313, line 157: Were you able to check if we can make a similar change as in the pandas case here to remove the hacky code here?

axl1313 (commented):
looks good other than one question about the pyspark case

cleanlab_studio/studio/studio.py
```diff
@@ -127,34 +128,29 @@ def apply_corrections(self, cleanset_id: str, dataset: Any, keep_excluded: bool
...
             )
             final = both.withColumn(
                 "__cleanlab_final_label",
-                # XXX hacky, checks if label is none by hand
-                # instead, use original JSON, which uses null values where it's not specified
-                udf(lambda original, clean: original if check_none(clean) else clean)(
-                    both[label_column],
-                    "clean_label",
-                ),
+                when(col("clean_label").isNull(), col(label_column)).otherwise(col("clean_label"))
```
axl1313, line 153: Thanks for figuring this out! Much better than the original `udf` implementation. Only issue here is I believe should be using "corrected_label" instead of "clean_label".

axl1313:
Can you also fix the formatter issues that are failling? You should be able to get the `black` vscode extension and set up autoformatting on save so you don't have to manually reformat

axl1313 (commented):
Found one more bug while testing, but looks good otherwise

axl1313 (approved):
lgtm!

PR #145, opened 2023-11-08
Snowpark Integration for upload_dataset and apply_corrections

AnimatorJoe:
Adding support for `snowflake.snowpark.DataFrame` type in `upload_dataset` and `apply_corrections` api functions.

### DIFF
--- a/cleanlab_studio/internal/util.py
+++ b/cleanlab_studio/internal/util.py
@@ -5,6 +5,8 @@ import math
 import numpy as np
 import pandas as pd
 
+import snowflake.snowpark as snowpark
+
 try:
     import pyspark.sql
 
@@ -16,12 +18,13 @@ from .dataset_source import (
     DatasetSource,
     FilepathDatasetSource,
     PandasDatasetSource,
+    SnowparkDatasetSource,
 )
 
 dataset_source_types = (
-    Union[str, pathlib.Path, pd.DataFrame]
+    Union[str, pathlib.Path, pd.DataFrame, snowpark.DataFrame]
     if not pyspark_exists
-    else Union[str, pathlib.Path, pd.DataFrame, pyspark.sql.DataFrame]
+    else Union[str, pathlib.Path, pd.DataFrame, snowpark.DataFrame, pyspark.sql.DataFrame]
 )
 
 DatasetSourceType = TypeVar("DatasetSourceType", bound=dataset_source_types)  # type: ignore
@@ -40,6 +43,8 @@ def init_dataset_source(
         return FilepathDatasetSource(
             filepath=pathlib.Path(dataset_source), dataset_name=dataset_name
         )
+    elif isinstance(dataset_source, snowpark.DataFrame):
+        return SnowparkDatasetSource(df=dataset_source, dataset_name=dataset_name)
     elif pyspark_exists and isinstance(dataset_source, pyspark.sql.DataFrame):
         from .dataset_source import PySparkDatasetSource
 
--- a/cleanlab_studio/internal/util.py
+++ b/cleanlab_studio/internal/util.py
@@ -5,7 +5,12 @@ import math
 import numpy as np
 import pandas as pd
 
-import snowflake.snowpark as snowpark
+try:
+    import snowflake.snowpark as snowpark
+
+    snowpark_exists = True
+except ImportError:
+    snowpark_exists = False
 
 try:
     import pyspark.sql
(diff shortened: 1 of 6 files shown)


axl1313 (commented):
Looks great! Just need to add to [`mypy.ini`](https://github.com/cleanlab/cleanlab-studio/blob/main/.mypy.ini) so that mypy ignores the missing `snowpark` imports

axl1313:
Can you fix the formatter and type check failures?

cleanlab_studio/studio/studio.py
```diff
@@ -126,7 +134,60 @@ def apply_corrections(self, cleanset_id: str, dataset: Any, keep_excluded: bool
         project_id = api.get_project_of_cleanset(self._api_key, cleanset_id)
         label_column = api.get_label_column_of_project(self._api_key, project_id)
         id_col = api.get_id_column(self._api_key, cleanset_id)
-        if _pyspark_exists and isinstance(dataset, pyspark.sql.DataFrame):
+
+        if _snowflake_exists and isinstance(dataset, snowpark.DataFrame):
```
axl1313, line 138: Can we refactor the `apply_corrections()` method and pull out the code to handle the separate dataset types into individual helper methods to avoid having this function continue to grow out of control?

cleanlab_studio/studio/studio.py
```diff
@@ -65,7 +70,7 @@ def upload_dataset(
         Uploads a dataset to Cleanlab Studio.
 
         Args:
-            dataset: Object representing the dataset to upload. Currently supported formats include a `str` path to your dataset, a pandas DataFrame, a pyspark DataFrame.
+            dataset: Object representing the dataset to upload. Currently supported formats include a `str` path to your dataset, a pandas snowflake, or pyspark DataFrame.
```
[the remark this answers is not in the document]
AnimatorJoe: oops

cleanlab_studio/internal/util.py
```diff
@@ -50,6 +63,124 @@ def init_dataset_source(
...
+    new_labels = final.select([id_col, action_col, cleanlab_final_label_col]).withColumnRenamed(
+        cleanlab_final_label_col, label_col
+    )
+
+    corrected_ds = corrected_ds.drop(label_col).join(new_labels, on=id_col, how="left")
+    corrected_ds = (
+        corrected_ds.where((col(action_col) != "exclude") | is_null(col(action_col)))
+        if not keep_excluded
+        else corrected_ds
+    ).drop(action_col)
```
axl1313, line 114: I think we can clean this up a bit by just doing:

- select id_col, action_col, corrected_label_col from cl_cols
- left join with original dataset on id_col
- temporarily add final label col
- drop original label col and corrected_label_col
- rename final label col to original label col name
- filter out excluded rows
- drop action col

This removes the need to do the multiple joins.

cleanlab_studio/internal/util.py
```diff
@@ -63,3 +194,11 @@ def check_none(x: Any) -> bool:
...
 def check_not_none(x: Any) -> bool:
     return not check_none(x)
+
+
+def quote(s: str) -> str:
+    return f'"{s}"'
+
+
+def quote_list(l: list) -> list:
+    return list(map(quote, l))
```
axl1313, line 204: just a style thing, but we typically prefer using list comprehensions over `list(map(...))` since it's a little more readable

axl1313 (commented):
Code mostly looks good and seems to work well! Great work! Just have a couple small comments

cleanlab_studio/internal/util.py
```diff
@@ -50,6 +63,127 @@ def init_dataset_source(
...
+        .join(
+            corrected_ds,
+            on=id_col,
+            how="left",
+        )
+        .withColumn(
+            cleanlab_final_label_col,
+            when(is_null(corrected_label_col), col(label_col)).otherwise(col(corrected_label_col)),
+        )
+        .drop(label_col)
```
axl1313, line 108: I believe we should be able to just do `.drop(label_col, corrected_label_col)` here and get rid of the `.select()` at the end of this. Let me know if I'm missing something though.

AnimatorJoe: There are several (~15ish) other columns in `cl_cols` which is attained from `download_cleanlab_columns`.

axl1313: At the beginning of this query though, we do `cl_cols_snowflake.select([id_col, action_col, corrected_label_col])` before any of the joins, renaming, coalescing, etc, so we should already have gotten rid of those extraneous columns right?

axl1313 (approved):
lgtm!

PR #153, opened 2023-12-06
python api regression support

axl1313:


ryansingman (approved):
LGTM. reminder, has to be merged after training, backend

PR #158, opened 2023-12-18
Supporting Importing of Image Datasets from Snowflake Stages

AnimatorJoe:
Adding support for importing image datasets stored from Snowflake stages and DBFS following [simple layout](https://help.cleanlab.ai/guide/concepts/datasets/#simple-zip) and [metadata layout](https://help.cleanlab.ai/guide/concepts/datasets/#metadata-zip).


cleanlab_studio/utils/databricks_utils.py
```diff
@@ -0,0 +1,88 @@
...
+    https://help.cleanlab.ai/guide/concepts/datasets/#image
+
+    Args:
+        folder_path: The POSIX-style path to your imageset.
+        archive_name: A name you choose for the imageset archive.
+
+    Returns:
+        The relative path to the archived imageset.
+    """
+    folder_name = os.path.basename(os.path.normpath(folder_path))
```
axl1313, line 19: somewhat of a nit, but we generally prefer using `pathlib` over `os`. `pathlib` is the more modern library. if it's not too much effort, can you rewrite these utils using `pathlib`?

cleanlab_studio/utils/databricks_utils.py
```diff
@@ -0,0 +1,135 @@
...
+    Args:
+        dbfs_path: The DBFS path to convert.
+
+    Returns:
+        The POSIX path.
+    """
+    return f"/dbfs{dbfs_path[5:]}"
+
+
+def get_databricks_imageset_df_image_col(df) -> str:
```
axl1313, line 20: Should add typing to the `df` input here (and other places in `databricks_utils` and `snowflake_utils` where it's missing

cleanlab_studio/utils/databricks_utils.py
```diff
@@ -0,0 +1,135 @@
...
+        archive_name
+        if archive_name is not None
+        else f'{folder_name}_archive_{datetime.now().strftime("%Y-%d-%m_%H:%M:%S")}'
+    )
+    output_filename = f"{archive_name}.zip"
+
+    # Create a ZipFile object in write mode
+    with zipfile.ZipFile(output_filename, "w", zipfile.ZIP_DEFLATED) as zipf:
+        # Walk through the directory
+        for root, _, files in os.walk(folder_path):
```
axl1313, line 68: I believe `pathlib.Path().walk()` should be roughly equivalent to `os.walk()` https://docs.python.org/3/library/pathlib.html#pathlib.Path.walk

cleanlab_studio/utils/databricks_utils.py
```diff
@@ -0,0 +1,135 @@
...
+            row = row.asDict()
+            original_path = dbfs_to_posix_path(row[image_col].origin)
+            path_in_zip = (
+                Path(archive_name)
+                .joinpath(Path(original_path).resolve(strict=False).name)
+                .as_posix()
+            )
+
+            # add row image to the zip file
+            zipf.write(original_path, arcname=path_in_zip)
```
axl1313, line 133: Is it possible to write the images to the zipfile in the same iteration as creating the metadata_file rather than iterating through the dataframe twice? Especially given that the first few lines of code in each iteration are repeated

AnimatorJoe: No, because I get the error ``ValueError: Can't write to ZIP archive while an open writing handle exists``.

cleanlab_studio/utils/databricks_utils.py
AnimatorJoe: Doesn't seem to be included in Databricks' default env.
Screen Shot 2024-01-25 at 10 57 58 PM

axl1313: Oh I think it's new in python 3.12. Probably best not to use then since we don't want to force people to be on latest python version

axl1313: It should be possible to do this with other pathlib methods, but since we're low on time this is fine

cleanlab_studio/utils/databricks_utils.py
```diff
@@ -104,12 +105,10 @@ def create_df_based_imageset_archive(df, archive_name=None) -> str:
...
         for row in df.toLocalIterator():
             row = row.asDict()
-            original_path = dbfs_to_posix_path(row[image_col].origin)
+            original_path = dbfs_to_posix_path(Path(row[image_col].origin))
             del row[image_col]
             path_in_zip = (
-                Path(archive_name)
-                .joinpath(Path(original_path).resolve(strict=False).name)
-                .as_posix()
+                Path(archive_name).joinpath(original_path.resolve(strict=False).name).as_posix()
```
axl1313, line 111: nit: can we just initialize `archive_name` as a `pathlib.Path()` instead of converting it all these separate times?

cleanlab_studio/utils/snowflake_utils.py
```diff
@@ -0,0 +1,111 @@
+import os
+import pandas as pd
+
+
+def get_snowflake_datarows(snowflake_cursor, stage_name, signed_url_expiration=604800):
```
axl1313, line 5: Add type annotations in snowflake_utils as well

axl1313 (commented):
Mostly looks good now. Can you fix the formatter errors?

axl1313:
Also fix CI errors https://github.com/cleanlab/cleanlab-studio/actions/runs/7665091436/job/20890396198?pr=158

cleanlab_studio/utils/databricks_utils.py
AnimatorJoe: `archive_name` should just be the name of the zip. Makes more sense to keep as `str`.

cleanlab_studio/utils/databricks_utils.py
```diff
@@ -0,0 +1,131 @@
...
+            row["filepath"] = path_in_zip
+
+            # write row to a csv file called metadata.csv
+            if first_row:
+                metadata_file.write((",".join(map(str, row.keys())) + "\n").encode("utf-8"))
+                first_row = False
+            metadata_file.write((",".join(map(str, row.values())) + "\n").encode("utf-8"))
+        metadata_file.close()
+
+        for row in df.toLocalIterator():
```
axl1313, line 121: Can we add a progress indicator here? It takes quite a while to run even on a dataset w only 300 images

PR #161, opened 2023-12-27
handle NA values for multilabel inference

axl1313:
Currently get errors when running multi-label inference if a row has no suggested labels. This PR adds handling for this edge case.

Screenshot 2023-12-27 at 1 50 45 PM


clu0 (approved):
lg

PR #166, opened 2024-01-09
better error message when id column doesn't exist

clu0:
Fixes this: https://github.com/cleanlab/cleanlab-studio-integration/issues/1431

### DIFF
--- a/cleanlab_studio/internal/upload_helpers.py
+++ b/cleanlab_studio/internal/upload_helpers.py
@@ -39,6 +39,10 @@ def upload_dataset(
         if modality is not None:
             schema["metadata"]["modality"] = modality
         if id_column is not None:
+            if id_column not in schema["fields"]:
+                raise ValueError(
+                    f"ID column {id_column} not found in dataset columns: {list(schema['fields'].keys())}"
+                )
             schema["metadata"]["id_column"] = id_column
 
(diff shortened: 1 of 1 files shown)


cleanlab_studio/internal/upload_helpers.py
```diff
@@ -39,6 +39,10 @@ def upload_dataset(
         if modality is not None:
             schema["metadata"]["modality"] = modality
         if id_column is not None:
+            if id_column not in schema["fields"]:
```
axl1313, line 42: should we also add this check for `label_column`, `text_column`, etc.?

clu0: During upload, it seems that only the `id_column` is used, because other columns are inferred.

During project creation, label column seems already to be handled [here](https://github.com/cleanlab/cleanlab-studio/blob/a7aebfb3da81b62b857dd88a480151ee64639cca/cleanlab_studio/studio/studio.py#L193), text column [here](https://github.com/cleanlab/cleanlab-studio/blob/a7aebfb3da81b62b857dd88a480151ee64639cca/cleanlab_studio/studio/studio.py#L212), and that code block also handled a few other cases.

axl1313: Ohh idk why i thought this was the project creation code

PR #168, opened 2024-01-18
bump model inference time limit

clu0:
As title, bump time limit to slightly larger than lambdas time limit

### DIFF
--- a/cleanlab_studio/studio/inference.py
+++ b/cleanlab_studio/studio/inference.py
@@ -39,7 +39,7 @@ class Model(abc.ABC):
         self,
         batch: Batch,
         return_pred_proba: bool = False,
-        timeout: int = 930,
+        timeout: int = 930,  # default is 15 mins (lambdas time limit) + 30s
     ) -> Union[Predictions, Tuple[Predictions, ClassProbablities]]:
         """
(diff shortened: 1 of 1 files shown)


cleanlab_studio/studio/inference.py
```diff
@@ -48,7 +48,7 @@ def predict(
         Args:
             batch: batch of examples to predict classes for
             return_pred_proba: whether to return predicted class probabilities for each example
-            timeout: optional parameter to set timeout for predictions in seconds (defaults to 600s)
+            timeout: optional parameter to set timeout for predictions in seconds (defaults to 930s, which is 15 mins for lambda time limit + 30s)
```
axl1313, line 51: not sure if we to explicitly mention that we're using lambdas here since this appears in the public docs

axl1313 (approved):
looks good other than one small comment about the docstring

cleanlab_studio/studio/inference.py
anishathalye: Yes, no need to explain implementation details in docs (users shouldn't care), would be good to explain the weird number in a code comment though (so it doesn't appear on help.cleanlab.ai).

PR #170, opened 2024-01-23
Hotfix: only apply nest_asyncio if in notebook

ryansingman:
Applying nest_asyncio while not in Jupyter notebook results in event loop locking. This change makes it such that we only apply nest_asyncio when in a jupyter notebook

Testing:
Run TLM for script or ipython session and from jupyter notebook


ryansingman:
https://github.com/tqdm/tqdm/blob/master/tqdm/autonotebook.py

inspiration for the `is_notebook` code ^^ (TQDM has to do a similar thing to us)

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -276,3 +280,13 @@ async def get_confidence_score_async(
...
+
+
+def is_notebook() -> bool:
+    """Returns True if running in a notebook, False otherwise."""
+    try:
+        get_ipython = sys.modules["IPython"].get_ipython
+        if "IPKernelApp" in get_ipython().config:
+            return True
+    finally:
+        return False
```
axl1313, line 292: Seems like this always returns False because of the `finally` block

ryansingman: 🤲 

good catch. shouldn't have tried to clean up the method w/o testing again 😢

axl1313 (approved):
I wasn't able to reproduce the issue before the change, but it's working well for me in jupyter notebook, interpreter, and script after the change

PR #173, opened 2024-01-25
Feature: add poll dataset ID for name method

ryansingman:
Adds poll dataset ID for name method to studio api


axl1313 (commented):
Mostly looks good. Just wondering what we want to happen if a dataset upload fails/is incomplete. I don't have context on the use case for this feature, so just want to double check that how we're handling this case is okay.

axl1313 (approved):

PR #180, opened 2024-02-12
pin black version

clu0:
as title, so we don't have to keep things updated with stable version


clu0:
pinning version doesn't seem to fix issues, need further investigation

PR #181, opened 2024-02-13
print link to cleanset when ready

axl1313:
This PR updates the `wait_until_cleanset_ready()` method to print a link to the cleanset page when ready to make it easier for a user to navigate to their cleanset

### DIFF
--- a/cleanlab_studio/internal/clean_helpers.py
+++ b/cleanlab_studio/internal/clean_helpers.py
@@ -8,7 +8,12 @@ from cleanlab_studio.errors import CleansetError
 from cleanlab_studio.internal.api import api
 
 
-def poll_cleanset_status(api_key: str, cleanset_id: str, timeout: Optional[float] = None) -> None:
+def poll_cleanset_status(
+    api_key: str,
+    cleanset_id: str,
+    timeout: Optional[float] = None,
+    show_cleanset_link: bool = False,
+) -> None:
     start_time = time.time()
     res = api.get_cleanset_status(api_key, cleanset_id)
     spinner = itertools.cycle("|/-\\")
@@ -37,7 +42,12 @@ def poll_cleanset_status(api_key: str, cleanset_id: str, timeout: Optional[float
 
         if res["is_ready"]:
             pbar.update(pbar.total - pbar.n)
-            pbar.set_postfix_str(res["step_description"])
+            ready_description = res["step_description"]
+            if show_cleanset_link:
+                ready_description += (
+                    f" View your cleanset at: https://app.cleanlab.ai/cleansets/{cleanset_id}"
+                )
+            pbar.set_postfix_str(ready_description)
             return
 
         if res["has_error"]:
--- a/cleanlab_studio/studio/studio.py
+++ b/cleanlab_studio/studio/studio.py
@@ -1,6 +1,7 @@
 """
 Python API for Cleanlab Studio.
 """
+
 from typing import Any, List, Literal, Optional, Union
 from types import FunctionType
 import warnings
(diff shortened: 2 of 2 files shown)


ryansingman:
Does this print belong in the open source? `wait_until_cleanset_ready` will be used in use cases outside of in Jupyter notebooks, and it's generally best practice to avoid extraneous prints

axl1313:
> Does this print belong in the open source? `wait_until_cleanset_ready` will be used in use cases outside of in Jupyter notebooks, and it's generally best practice to avoid extraneous prints

Ah that's a good point. I guess when running in a script this print is unlikely to be useful. What if we do a check if running in Jupyter notebook and print if so?

ryansingman:
> > Does this print belong in the open source? `wait_until_cleanset_ready` will be used in use cases outside of in Jupyter notebooks, and it's generally best practice to avoid extraneous prints
> 
> Ah that's a good point. I guess when running in a script this print is unlikely to be useful. What if we do a check if running in Jupyter notebook and print if so?

That's a bit better, but I'm still not sure why this code should go in `cleanlab-studio` as opposed to in the notebooks themselves. Is there a reason to do this here?

axl1313:
> > > Does this print belong in the open source? `wait_until_cleanset_ready` will be used in use cases outside of in Jupyter notebooks, and it's generally best practice to avoid extraneous prints
> > 
> > 
> > Ah that's a good point. I guess when running in a script this print is unlikely to be useful. What if we do a check if running in Jupyter notebook and print if so?
> 
> That's a bit better, but I'm still not sure why this code should go in `cleanlab-studio` as opposed to in the notebooks themselves. Is there a reason to do this here?

I just figured it'd be useful for a lot of cases since you have to open the project in the web UI to do things like make corrections, rerun cleanlab, deploy models and this way we wouldn't need to update all the tutorials separately. What if we updated the [postfix str](https://github.com/cleanlab/cleanlab-studio/blob/47902969e14c2145d67e1d9cb9f8b751a2c926d7/cleanlab_studio/internal/clean_helpers.py#L40) on the tqdm progress to include the cleanset link when the cleanset's ready instead of printing a separate message?

ryansingman:
> > > > Does this print belong in the open source? `wait_until_cleanset_ready` will be used in use cases outside of in Jupyter notebooks, and it's generally best practice to avoid extraneous prints
> > > 
> > > 
> > > Ah that's a good point. I guess when running in a script this print is unlikely to be useful. What if we do a check if running in Jupyter notebook and print if so?
> > 
> > 
> > That's a bit better, but I'm still not sure why this code should go in `cleanlab-studio` as opposed to in the notebooks themselves. Is there a reason to do this here?
> 
> I just figured it'd be useful for a lot of cases since you have to open the project in the web UI to do things like make corrections, rerun cleanlab, deploy models and this way we wouldn't need to update all the tutorials separately. What if we updated the [postfix str](https://github.com/cleanlab/cleanlab-studio/blob/47902969e14c2145d67e1d9cb9f8b751a2c926d7/cleanlab_studio/internal/clean_helpers.py#L40) on the tqdm progress to include the cleanset link when the cleanset's ready instead of printing a separate message?

My general feeling here is that explicit > implicit. I do definitely agree that this will be useful in a lot of cases, but I think that that's only in tutorials and jupyter notebooks. How would a solution like adding a parameter (`show_cleanset_link`, default `False`) to the method feel to you? I like that it's more explicit and also prevents us from printing out links in the console for API users (which I think is clunky -- it gives the impression that the app is only usable via the UI)

ryansingman (approved):

PR #184, opened 2024-02-20
add deprecation warnings and schema override conversion

axl1313:
- Adds deprecation warnings if a user provides `id_column` or `modality` kwargs to the `upload_dataset` method
- Adds a deprecation warning if a user provides `schema_overrides` in the old format
- Attempts to convert old `schema_overrides` format to new format
  - If feature_type is "datetime" or "identifier", raises error
  - If data_type, feature_type combination is invalid in old schema format, raises error

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1206603310631125

### DIFF
--- a/cleanlab_studio/errors.py
+++ b/cleanlab_studio/errors.py
@@ -38,3 +38,7 @@ class InternalError(Exception):
 
 class CleansetError(InternalError):
     pass
+
+
+class InvalidSchemaTypeError(ValueError):
+    pass
--- a/cleanlab_studio/errors.py
+++ b/cleanlab_studio/errors.py
@@ -41,4 +41,8 @@ class CleansetError(InternalError):
 
 
 class InvalidSchemaTypeError(ValueError):
-    pass
+    def __init__(self, msg: str) -> None:
+        self.msg = msg
+
+    def __str__(self) -> str:
+        return f"{self.msg}\nSee [TODO: insert docs] link for more information."
--- a/cleanlab_studio/studio/studio.py
+++ b/cleanlab_studio/studio/studio.py
@@ -83,10 +83,12 @@ class Studio:
         if kwargs.get("modality") is not None:
             warnings.warn(
                 "Ignoring `modality` parameter which is deprecated and will be removed in a future release.",
+                FutureWarning,
             )
         if kwargs.get("id_column") is not None:
             warnings.warn(
                 "Ignoring `id_column` parameter which is deprecated and will be removed in a future release.",
+                FutureWarning,
             )
 
         if isinstance(schema_overrides, dict):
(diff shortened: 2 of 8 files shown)


cleanlab_studio/studio/studio.py
```diff
@@ -30,6 +31,8 @@
 if _pyspark_exists:
     import pyspark.sql
 
+warnings.simplefilter("always", DeprecationWarning)
```
ryansingman, line 34: question: is there a reason to set this here? we generally wouldn't want to override a users warning filter configureation

cleanlab_studio/studio/studio.py
```diff
@@ -63,21 +66,41 @@ def upload_dataset(
...
+                "Ignoring `id_column` parameter which is deprecated and will be removed in a future release.",
+                DeprecationWarning,
+            )
+
+        if isinstance(schema_overrides, dict):
+            # TODO: link to documentation for schema override format
+            warnings.warn(
+                "Using deprecated `schema_overrides` format. Please use list of SchemaOverride objects instead.",
+                DeprecationWarning,
+            )
```
ryansingman, line 101: should warn _after_ converting schema overrides to prevent case where we warn and then fail

cleanlab_studio/errors.py
```diff
@@ -38,3 +38,7 @@ class InternalError(Exception):
 
 class CleansetError(InternalError):
     pass
+
+
+class InvalidSchemaTypeError(ValueError):
```
ryansingman, line 43: these warnings should all show a link to documentation.

would suggest putting that docs link in the `__str__` method of this class. can put a todo for now since we haven't yet written docs 😢

ryansingman (commented):
couple small comments, otherwise LGTM!

cleanlab_studio/studio/studio.py
axl1313: DeprecationWarning is filtered out by default usually. I can switch the warnings to the default UserWarning instead and remove this override. (see [docs](https://docs.python.org/3/library/warnings.html#warning-categories))

ryansingman: https://docs.python.org/3/library/warnings.html#warning-categories

Ah looks like we can use `FutureWarning`,  which is both not ignored by default and is a better fit for what's happening here (we're not deprecating schema overrides, we're changing their behavior)

ryansingman (approved):
one suggestion to use `FutureWarning` (in thread), otherwise LGTM!

we can merge this into the `feature/upload-v1` branch w/ the TODOs for docs. We'll handle those before we merge `feature/upload-v1`

PR #185, opened 2024-02-21
[TLM upgrade] Add model type to initi params

ulya-tkch:
This PR changes how we handle the options dictionary w..r.t. parameters set during initialization and introduces a new parameter that can be set during initialization: `model`. 

How we handle options dictionary for a specific key in options dictionary that has a matching parameter that could be set during initialization:
- if key is passed in during prompt, it overrides that param set at initialization
- if no param set during initialization or passed into options dictionary, defaults values used
- otherwise, use param set during initialization


huiwengoh:
Closing in favor of updated API in #197

PR #186, opened 2024-02-21
[upload v1] support schema overrides during initial upload

axl1313:
- include schema overrides as part of confirm upload instead of requiring additional schema ingestion
- check for error status when polling ingestion progress

Test with [associated backend changes](https://github.com/cleanlab/cleanlab-studio-backend/pull/1211)


ryansingman (approved):
LGTM

ryansingman:
Wrong base ref for this branch? should be `feature/upload-v1` I believe

axl1313:
> Wrong base ref for this branch? should be `feature/upload-v1` I believe

Oops yes good catch. Forgot to update after merging the previous branch into `feature/upload-v1`

PR #190, opened 2024-02-27
Fix - populate final label in apply_corrections_pd_df with fillna

LukeMainwaring:
Bug fix:
Currently when the `apply_corrections_pd_df` function is called where the label column is of type integer, then it'll throw `TypeError: only integer scalar arrays can be converted to a scalar index`. It appears this error is due to the use of `.tolist()` method on a pandas Series that is being used as an indexer. In pandas, when you use a Series or a list as an indexer, it expects the list or Series to be of boolean type, not integer. 

solution:
The `where` function here is used to replace values where the condition is `False`, because we are replacing null values in the `corrected_label` column with the corresponding values from the `label_column` of the dataset. We can achieve this instead by using the `fillna` function.


axl1313 (approved):
lgtm!

PR #197, opened 2024-03-06
Update TLM API

huiwengoh:
Key changes:
- removes arguments `max_concurrent_request`, `retries`, `max_timeout`
- move some arguments to be set during initialization, such as `options`, `timeout`
  - adds a new argument `verbose`
- allow batching for async methods
- rename `get_confidence_score` -> `get_trustworthiness_score`

### DIFF
--- a/cleanlab_studio/internal/api/api.py
+++ b/cleanlab_studio/internal/api/api.py
@@ -511,7 +511,6 @@ async def tlm_prompt(
     prompt: str,
     quality_preset: str,
     options: Optional[JSONDict],
-    client_session: Optional[aiohttp.ClientSession] = None,
 ) -> JSONDict:
     """
     Prompt Trustworthy Language Model with a question, and get back its answer along with a confidence score
@@ -526,10 +525,7 @@ async def tlm_prompt(
     Returns:
         JSONDict: dictionary with TLM response and confidence score
     """
-    local_scoped_client = False
-    if not client_session:
-        client_session = aiohttp.ClientSession()
-        local_scoped_client = True
+    client_session = aiohttp.ClientSession()
 
     try:
         res = await client_session.post(
@@ -544,8 +540,7 @@ async def tlm_prompt(
         handle_api_error_from_json(res_json)
 
     finally:
-        if local_scoped_client:
-            await client_session.close()
+        await client_session.close()
 
     return cast(JSONDict, res_json)
 
@@ -557,7 +552,6 @@ async def tlm_get_confidence_score(
     response: str,
     quality_preset: str,
     options: Optional[JSONDict],
-    client_session: Optional[aiohttp.ClientSession] = None,
 ) -> JSONDict:
     """
     Query Trustworthy Language Model for a confidence score for the prompt-response pair.
@@ -568,15 +562,11 @@ async def tlm_get_confidence_score(
         response (str): response for TLM to get confidence score for
         quality_preset (str): quality preset to use to generate confidence score
         options (JSONDict): additional parameters for TLM
-        client_session (aiohttp.ClientSession): client session used to issue TLM request
 
     Returns:
         JSONDict: dictionary with TLM confidence score
     """
-    local_scoped_client = False
-    if not client_session:
-        client_session = aiohttp.ClientSession()
-        local_scoped_client = True
+    client_session = aiohttp.ClientSession()
 
     try:
         res = await client_session.post(
@@ -588,15 +578,11 @@ async def tlm_get_confidence_score(
         )
         res_json = await res.json()
 
-        if local_scoped_client:
-            await client_session.close()
-
         await handle_tlm_client_error_from_resp(res)
         handle_rate_limit_error_from_resp(res)
         handle_api_error_from_json(res_json)
 
     finally:
-        if local_scoped_client:
-            await client_session.close()
+        await client_session.close()
 
     return cast(JSONDict, res_json)
--- a/cleanlab_studio/internal/constants.py
+++ b/cleanlab_studio/internal/constants.py
@@ -2,5 +2,5 @@ from typing import List
 
 # prepend constants with _ so that they don't show up in help.cleanlab.ai docs
 _DEFAULT_MAX_CONCURRENT_TLM_REQUESTS: int = 16
-_MAX_CONCURRENT_TLM_REQUESTS_LIMIT: int = 128
 _VALID_TLM_QUALITY_PRESETS: List[str] = ["best", "high", "medium", "low", "base"]
+_TLM_MAX_RETRIES: int = 10  # TODO: finalize this number
--- a/cleanlab_studio/studio/trustworthy_language_model.py
+++ b/cleanlab_studio/studio/trustworthy_language_model.py
@@ -124,7 +124,7 @@ class TLM:
 
     def _batch_prompt(
         self,
-        prompts: List[str],
+        prompts: Sequence[str],
     ) -> List[TLMResponse]:
         """Run batch of TLM prompts.
 
@@ -142,14 +142,14 @@ class TLM:
 
     def _batch_get_trustworthiness_score(
         self,
-        prompts: List[str],
-        responses: List[str],
+        prompts: Sequence[str],
+        responses: Sequence[str],
     ) -> List[float]:
         """Run batch of TLM get confidence score.
 
         Args:
-            prompts (List[str]): list of prompts to run get confidence score for
-            responses (List[str]): list of responses to run get confidence score for
+            prompts (Sequence[str]): list of prompts to run get confidence score for
+            responses (Sequence[str]): list of responses to run get confidence score for
 
         Returns:
             List[float]: TLM confidence score for each prompt (in supplied order)
@@ -225,7 +225,9 @@ class TLM:
         if isinstance(prompt, str):
             return await self._prompt_async(prompt)
 
-        return await self._batch_async([self._prompt_async(p) for p in prompt])
+        return cast(
+            List[TLMResponse], await self._batch_async([self._prompt_async(p) for p in prompt])
+        )
 
     async def _prompt_async(
         self,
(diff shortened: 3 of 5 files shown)


ryansingman (changes requested):

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -72,119 +69,97 @@ def __init__(
...
+                "See: https://help.cleanlab.ai/reference/python/trustworthy_language_model/#class-tlmoptions"
+            )
+
+        self._options = options
+
+        if timeout is not None:
+            self._timeout = None if timeout <= 0 else timeout
+        else:
+            # TODO: figure out how to compute appropriate timeout
+            self._timeout = 3600
```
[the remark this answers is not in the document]
huiwengoh: Per the api spec:

> - `timeout` semantics
>     - If not specified, calculated automatically
>     - If `<= 0`, then infinite
>     - Otherwise: per-prompt timeout, in seconds

I'm not entirely sure what _calculated automatically_ means here, but my understanding was that we would adjust it (based on something)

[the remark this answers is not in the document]
ryansingman: Ok that's bizarre - I've never seen a timeout arg constructed like that.

Can you message Anish and ask for clarification there? I can't think of anything that would be a "reasonable timeout" here

cleanlab_studio/internal/api/api.py
```diff
@@ -526,10 +525,7 @@ async def tlm_prompt(
     Returns:
         JSONDict: dictionary with TLM response and confidence score
     """
-    local_scoped_client = False
-    if not client_session:
-        client_session = aiohttp.ClientSession()
-        local_scoped_client = True
```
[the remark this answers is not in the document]
huiwengoh: Updated, I'm not 100% sure if I'm doing it right tho so please advice otherwise!

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -195,184 +172,168 @@ async def _batch_async(
...
-            options (Optional[TLMOptions]): options to parameterize TLM with. Defaults to None.
-            client_session (Optional[aiohttp.ClientSession]): async HTTP session to use for TLM query. Defaults to None.
-            retries (int): number of retries for TLM query. Defaults to 0.
         Returns:
-            TLMResponse: [TLMResponse](#class-tlmresponse) object containing the response and confidence score
+            TLMResponse: [TLMResponse](#class-tlmresponse) object containing the response and trustworthiness score
         """
+
         async with self._query_semaphore:
             tlm_response = await api.tlm_prompt(
```
ryansingman, line 256: need to apply timeout here?

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -389,9 +350,9 @@ async def get_confidence_score_async(
                         prompt,
```
ryansingman, line 350: need to apply timeout here?

cleanlab_studio/internal/constants.py
```diff
@@ -1,6 +1,11 @@
...
+# TLM constats
 # prepend constants with _ so that they don't show up in help.cleanlab.ai docs
 _DEFAULT_MAX_CONCURRENT_TLM_REQUESTS: int = 16
-_MAX_CONCURRENT_TLM_REQUESTS_LIMIT: int = 128
 _VALID_TLM_QUALITY_PRESETS: List[str] = ["best", "high", "medium", "low", "base"]
+_VALID_TLM_MODELS: List[str] = ["gpt-3.5-turbo-16k", "gpt-4"]
+_TLM_MAX_RETRIES: int = 10  # TODO: finalize this number
+TLM_MAX_TOKEN_RANGE: Tuple[int, int] = (64, 512)  # (min, max)
+TLM_NUM_CANDIDATE_RESPONSES_RANGE: Tuple[int, int] = (1, 100)  # (min, max)
+TLM_NUM_CONSISTENCY_SAMPLES_RANGE: Tuple[int, int] = (0, 50)  # (min, max)
```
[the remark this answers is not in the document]
huiwengoh: If I'm not mistaken, it's only the constants that get imported to `trustworthy_language_model.py` that will appear in docs, this is only used in the `tlm_helpers.py` so it should be fine, I can double check

cleanlab_studio/studio/trustworthy_language_model.py
huiwengoh: Added timeout but not here, I don't think we can add it in this helper function bc the batch function calls this helper function and we want a overall timeout there (not on each individual prompt).

cleanlab_studio/studio/trustworthy_language_model.py
huiwengoh: Similar to comment above

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -72,307 +78,383 @@ def __init__(
...
-        else:
-            options = cast(Union[None, TLMOptions], options)
-            options_collection = [options for _ in prompts]
+        tlm_responses = self._event_loop.run_until_complete(
+            self._batch_async([self._prompt_async(prompt) for prompt in prompts])
+        )
+
+        return cast(List[TLMResponse], tlm_responses)
+
+    def _batch_prompt_with_handled_errors(
```
huiwengoh, line 147: It might be a little repetitive to define this function (could potentially combine with `_batch_prompt` with a `handle_errors` flag), but I was getting way too many mypy errors (which I wasn't sure how to handle/if I could ignore) so decided to do it separately.

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -72,307 +78,383 @@ def __init__(
...
-                timeout=timeout,
+                    self._get_trustworthiness_score_async(prompt, response)
+                    for prompt, response in zip(prompts, responses)
+                ]
             )
         )
 
         return cast(List[float], tlm_responses)
 
+    def _batch_get_trustworthiness_score_with_handled_errors(
```
huiwengoh, line 193: Same comment as `batch_prompt_with_handled_errors`

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -72,307 +78,383 @@ def __init__(
...
+            with tqdm(total=len(tlm_query_tasks)) as pbar:
+                for task in asyncio.as_completed(tlm_query_tasks):
+                    try:
+                        result = await asyncio.wait_for(task, timeout=self._timeout)
+                        results.append(result)
+                    except Exception:
+                        results.append(None)
+                    pbar.update(1)
+
+            return results  # type: ignore[return-value]
```
huiwengoh, line 259: An issue with this is that I'm sometimes getting this log:
```
Task exception was never retrieved
future: <Task finished name='Task-14' coro=<wait_for() done, defined at /Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/asyncio/tasks.py:421> exception=TimeoutError()>
```

when running something like
```
tlm = studio.TLM( timeout=0.1)
tlm.try_prompt(["what is 1+2", "what is 3435+2"])
```

I think that is happening because of the for loop, but I am not sure how else to use `tqdm` and return exceptions at the same time (`tqdm_asyncio.gather` does not have a `return_exceptions` argument)

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -72,307 +78,383 @@ def __init__(
...
+    ) -> Union[List[Optional[TLMResponse]], List[Optional[float]]]:
+        tlm_query_tasks = [
+            asyncio.wait_for(asyncio.create_task(tlm_coro), timeout=self._timeout)
+            for tlm_coro in tlm_coroutines
+        ]
+
+        if self._verbose:
+            results = []
+            with tqdm(total=len(tlm_query_tasks)) as pbar:
+                for task in asyncio.as_completed(tlm_query_tasks):
```
ryansingman, line 251: this will return responses out of order

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -72,307 +78,383 @@ def __init__(
...
+                    try:
+                        result = await asyncio.wait_for(task, timeout=self._timeout)
```
ryansingman, line 253: redoing a wait_for (likely causing the task exception never retrieved error)

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -72,307 +78,383 @@ def __init__(
...
             prompt (str): prompt for the TLM
-            options (Optional[TLMOptions]): options to parameterize TLM with. Defaults to None.
-            client_session (Optional[aiohttp.ClientSession]): async HTTP session to use for TLM query. Defaults to None.
-            retries (int): number of retries for TLM query. Defaults to 0.
         Returns:
-            TLMResponse: [TLMResponse](#class-tlmresponse) object containing the response and confidence score
+            TLMResponse: [TLMResponse](#class-tlmresponse) object containing the response and trustworthiness score
         """
+
         async with self._query_semaphore:
```
ryansingman, line 349: need to move semaphore / limiting to inside `tlm_prompt` (to properly handle retries)

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -72,307 +78,383 @@ def __init__(
...
-            options (Optional[TLMOptions]): options to parameterize TLM with. Defaults to None.
-            client_session (Optional[aiohttp.ClientSession]): async HTTP session to use for TLM query. Defaults to None.
-            retries (int): number of retries for TLM query. Defaults to 0.
+            response: response for the TLM to evaluate
+            client_session: async HTTP session to use for TLM query. Defaults to None.
         Returns:
-            float corresponding to the TLM's confidence score
+            float corresponding to the TLM's trustworthiness score
 
         """
```
ryansingman, line 459: as w/ _prompt_async, move semaphore / limiting into `tlm_get_confidence_score`

PR #193, opened 2024-02-29
Patch: dont retry tlm 4XXs

ryansingman:
Raise 4XXs from TLM immediately, don't retry

---

Testing:
Run the TLM, providing invalid options. TLM should fail ~immediately without retrying, with a `TlmBadRequest` error type:

E.g:
```
tlm = Studio("<API_KEY>").TLM()
tlm.prompt("2+2?", options={"some_invalid_option": 123})
```

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1206649294650263

PR #198, opened 2024-03-12
Hotfix: rate limit error precedence

ryansingman:
handle rate limits before tlm client errors
(ensure rate limits get retried

Fixes bug introduced https://github.com/cleanlab/cleanlab-studio/pull/193


huiwengoh (approved):
No longer getting token limit errors, thanks for the fix!

PR #202, opened 2024-03-14
base class for handled errors

clu0:
Create a base class for all handled errors, and add a `is_handled_error` attribute in logging.


LukeMainwaring (approved):

LukeMainwaring (changes requested):
code lgtm, but it looks like there are merge conflicts right now

LukeMainwaring (approved):

PR #204, opened 2024-03-22
Feature: TLM API

ryansingman:
- Simplifies TLM API (no more `batch_` methods
- Unifies TLM option setting
- Updates concurrency scheme

See also, previous PR (which contains some code reviews): https://github.com/cleanlab/cleanlab-studio/pull/197


anishathalye:
I've tweaked the docs in https://github.com/cleanlab/cleanlab-studio/pull/204/commits/0568bf4f298fa08292b404cd7f492e5e574fc7d7 (note I also got rid of some type aliases in that commit), docs look good on my end.

For getting this merged and deployed - do we need another round of code review or are we good to go? There where some reviews in the old PR (https://github.com/cleanlab/cleanlab-studio/pull/197) but not a final approval.

ryansingman:
> I've tweaked the docs in [0568bf4](https://github.com/cleanlab/cleanlab-studio/commit/0568bf4f298fa08292b404cd7f492e5e574fc7d7) (note I also got rid of some type aliases in that commit), docs look good on my end.
> 
> For getting this merged and deployed - do we need another round of code review or are we good to go? There where some reviews in the old PR (#197) but not a final approval.
> 
> Who is the right person to review this?

Another round of review would likely be beneficial for such a large change (though we've done incremental reviews on branches being merged into this one).

Luke would likely be the correct person to review

cleanlab_studio/internal/tlm/concurrency.py
```diff
@@ -0,0 +1,113 @@
...
+        congestion_window_increase = self._congestion_window - prev_congestion_window
+        for _ in range(congestion_window_increase):
+            self._send_semaphore.release()
+
+    async def _decrease_congestion_window(
+        self,
+        multiplicative_decrease_factor: int = MULTIPLICATIVE_DECREASE_FACTOR,
+    ) -> None:
+        """Decreases TLM congestion window, to minimum of 1."""
+        if self._congestion_window == 1:
```
LukeMainwaring, line 104: nit: would it be safer to set this to `if self._congestion_window <= 1:` ? it wouldn't happen with the current constants, but technically `self. _congestion_window` could go to 0 with other argument values passed in

cleanlab_studio/internal/tlm/concurrency.py
```diff
@@ -0,0 +1,113 @@
...
+            return
+
+        prev_congestion_window = self._congestion_window
+        self._congestion_window //= multiplicative_decrease_factor
+
+        # acquire congestion window decrease from send semaphore
+        congestion_window_decrease = prev_congestion_window - self._congestion_window
+        for _ in range(congestion_window_decrease):
+            await self._send_semaphore.acquire()
```
LukeMainwaring, line 113: i'm not aware how likely it is to have issues acquiring the semaphore, but if so then this could lead to a deadlock situation. instead of letting the function block indefinitely, we could set a timeout. not a super strong opinion here and maybe something we add later if it's actually an issue we encounter

  ```
try:
        await asyncio.wait_for(self._send_semaphore.acquire(), timeout=0.1)
    except asyncio.TimeoutError:
        break
```

File: cleanlab_studio/internal/tlm/validation.py [hunk 3]
```diff
@@ -0,0 +1,178 @@
...
+from cleanlab_studio.errors import ValidationError
+from cleanlab_studio.internal.constants import (
+    _VALID_TLM_MODELS,
+    TLM_MAX_TOKEN_RANGE,
+    TLM_NUM_CANDIDATE_RESPONSES_RANGE,
+    TLM_NUM_CONSISTENCY_SAMPLES_RANGE,
+)
+
+
+SKIP_VALIDATE_TLM_OPTIONS = os.environ.get("CLEANLAB_STUDIO_SKIP_VALIDATE_TLM_OPTIONS", False)
```
LukeMainwaring, line 12: nit: could be more explicit that this is a boolean flag. the env var will be read as a string and could lead to weird behavior "false" evaluating to True because non-empty strings are truthy. 

one potential option:
`SKIP_VALIDATE_TLM_OPTIONS: bool = os.environ.get("CLEANLAB_STUDIO_SKIP_VALIDATE_TLM_OPTIONS", "false").lower() == "true"`

cleanlab_studio/internal/tlm/validation.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+def validate_tlm_prompt(prompt: Union[str, Sequence[str]]) -> None:
+    if isinstance(prompt, str):
+        return
+
+    elif isinstance(prompt, Sequence):
```
LukeMainwaring, line 19: imo error messages would be a little more helpful if they were more specific, such as pointing out the incorrect item type like below
```
    elif isinstance(prompt, Sequence):
        for item in prompt:
            if not isinstance(item, str):
                raise ValidationError(
                    f"Item {item} in prompt is of invalid type {type(item)}, all items in the prompt list must be of type str."
                )
```

cleanlab_studio/internal/tlm/validation.py
```diff
@@ -0,0 +1,178 @@
...
+                "Some items in prompt are of invalid types, all items in the prompt list must be of type str."
+            )
+
+    else:
+        raise ValidationError(
+            f"Invalid type {type(prompt)}, prompt must either be strings or list/iterable of strings."
+        )
+
+
+def validate_tlm_try_prompt(prompt: Sequence[str]) -> None:
```
LukeMainwaring, line 31: not a blocker for this PR, but there is a decent amount of code duplication in this file. if we plan on adding more types of queries than prompt and response in the future, then it'll be easier to maintain with a helper function that checks type(s) of string/sequence as string.

LukeMainwaring:
left some comments but nothing blocking for release. feel free to reply to these comments or via dm, whichever is easiest

cleanlab_studio/internal/tlm/concurrency.py
ryansingman: the semaphore gets released every time a request finishes or errors, so it _should not_ be possible to have a deadlock here

I don't think it makes sense to add that handling code here -- it likely means that something else catastrophically broke. further, there's no clear value to set the timeout to.

cleanlab_studio/internal/tlm/validation.py
ryansingman: we would want to point out all invalid indexes for this type of validation, which is slightly less trivial (especially considering the code duplication you pointed out)

This would be a non-breaking future API change, so we will defer this.

PR #205, opened 2024-03-27
add better validation for project label column

axl1313:
### Description

This PR contains a few updates to label column validation when creating projects through the Python API:
- Improved error message if label column does not have valid type. Type validation is done in backend in dataset details endpoint.
- Adds column diversity validation for multi-class projects
- Adds validation for multi-label label columns

More context in [notion](https://www.notion.so/cleanlab/Label-column-validation-0fe0f7897f9340288859bcf3f65e6cca?pvs=4)

Note: must be merged after [backend changes](https://github.com/cleanlab/cleanlab-studio-backend/pull/1373)

### How to test

Test w/ [backend changes](https://github.com/cleanlab/cleanlab-studio-backend/pull/1373)
- Attempt to create multi-class tabular project w/ label column containing < 2 classes w/ >= 5 examples each. This should fail.
- Attempt to create multi-class image/text project w/ label column containing < 2 classes w/ >= 5 examples each. This should succeed since it falls under the data labeling workflow
- Attempt to create multi-label project w/ label column that does not contain any comma separated string values. This should fail.
- Verify that error messages for the above are reasonably informative.

### DIFF
--- a/cleanlab_studio/internal/clean_helpers.py
+++ b/cleanlab_studio/internal/clean_helpers.py
@@ -1,10 +1,10 @@
 import itertools
 import time
-from typing import Optional
+from typing import List, Optional
 
 from tqdm import tqdm
 
-from cleanlab_studio.errors import CleansetError
+from cleanlab_studio.errors import CleansetError, InvalidDatasetError
 from cleanlab_studio.internal.api import api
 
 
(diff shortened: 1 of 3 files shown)


cleanlab_studio/internal/clean_helpers.py
```diff
@@ -53,3 +53,44 @@ def poll_cleanset_status(
         if res["has_error"]:
             pbar.set_postfix_str(res["step_description"])
             raise CleansetError(f"Cleanset {cleanset_id} failed to complete")
+
+
+def validate_label_column(
```
ryansingman, line 58: this is acceptable for now, but note that this is not a scalable way to handle validation. for every additional validation check that we add, we'll have to add an API method + a call to it from our client.

instead, we should centralize these validation methods (a `validate_project` endpoint) that contains all of the logic. then we can update that as desired without needing to make corresponding client changes every time.

Can you create a task to do so? this was the original intention we had when deciding to add matching validation to the Python client, but it wasn't specified for you here. I don't want to block this PR longer than necessary, but we should make this change later on

ryansingman (approved):
Changes LGTM -- just need to create task like I mentioned in review comment

cleanlab_studio/internal/clean_helpers.py
axl1313: Yeah that's a good point. Created a task [here](https://www.notion.so/cleanlab/Move-project-creation-validation-to-backend-ee78e332554644e0a436b2daee610290?pvs=4)

PR #208, opened 2024-03-29
add error handling for cleanset errors

axl1313:
This PR updates the `wait_until_cleanset_ready()` method to display the `error_type` and `error_message` which is now set on `Cleanset` records when an error occurs during training.


cleanlab_studio/internal/clean_helpers.py
```diff
@@ -52,4 +52,13 @@ def poll_cleanset_status(
 
         if res["has_error"]:
             pbar.set_postfix_str(res["step_description"])
+            error_message = res["error_message"]
+            error_type = res["error_type"]
+
+            if error_type is not None:
+                assert (
+                    error_message is not None
+                ), "Error message should be present for handled error"
```
ryansingman, line 61: if `error_message` isn't present for some reason, we wouldn't want to show this opaque assertion error to a user. we'd instead want to show them a default error message and send this telemetry back to the backend

ryansingman (changes requested):

ryansingman (approved):
LGTM! thanks for making changes

PR #210, opened 2024-04-02
Fix variables showing up in docs as globals

anishathalye:
This patch fixes "name" and "method" showing up in the list of global variables defined by this module (which was being rendered in the docs).

There are some other preexisting problems with the telemetry, including missing telemetry for the Model and TLM classes; this patch does not address those issues.

Before:
Screenshot 2024-04-02 at 7 37 54 PM

After:
Screenshot 2024-04-02 at 7 37 50 PM

### DIFF
--- a/cleanlab_studio/studio/studio.py
+++ b/cleanlab_studio/studio/studio.py
@@ -436,6 +436,9 @@ class Studio:
 
 
 # decorate all functions of self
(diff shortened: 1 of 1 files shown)


ulya-tkch (approved):
lgtm var names changed eveywhere they are used

PR #211, opened 2024-04-03
Clean up model docs

anishathalye:
- Explain how a Model is meant to be instantiated
- Hide `csv_string_to_list` from public API

### DIFF
--- a/cleanlab_studio/studio/inference.py
+++ b/cleanlab_studio/studio/inference.py
@@ -1,5 +1,9 @@
 """
-Methods for interfacing with deployed ML models (to produce predictions)
+Methods for interfacing with deployed ML models (to produce predictions).
+
+**This module is not meant to be imported and used directly.** Instead, use [`Studio.get_model()`](../studio/#method-get_model) to instantiate a [Model](#class-model) object.
+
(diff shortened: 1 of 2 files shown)


cleanlab_studio/studio/inference.py
```diff
@@ -26,10 +30,15 @@
...
+    """Represents a machine learning model instance in a Cleanlab Studio account.
+
+    Models should be instantiated using the [`Studio.get_model()`](../studio/#method-get_model) method. Then, using a Model object, you can [`predict()`](#method-predict) labels for new data.
+    """
 
     def __init__(self, api_key: str, model_id: str):
-        """Initializes model class w/ API key and model ID."""
+        """Initializes a model.
+
+        **Objects of this class are not meant to be constructed directly.** Instead, use [`Studio.get_model()`](../studio/#method-get_model)."""
```
ulya-tkch, line 41: *add why you'd use get_model()

```suggestion
        **Objects of this class are not meant to be constructed directly.** Instead, use [`Studio.get_model()`](../studio/#method-get_model) to get a Model object."""
```

ulya-tkch (approved):

PR #213, opened 2024-04-07
fix dataframe index when downloading cleanlab columns

axl1313:
### Description
This PR fixes an issue that could occur when merging cleanlab columns with the original dataset. After sorting cleanlab columns by id column, it was possible for the index of the dataframe to no longer be in sorted order. Adding the [`ignore_index` parameter](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.sort_values.html) when sorting the dataframe ensures that the resulting dataframe's index column is of the form 0, 1, ... n-1 so that we can merge with the original dataset using `df.merge(cleanlab_columns, left_index=True, right_index=True)` properly.

### How to test
- Tabular quickstart tutorial is an example of a tutorial where merging results was broken before this change. Can run this tutorial and check that results are the same as in the Web UI.


ryansingman (approved):
LGTM

PR #231, opened 2024-05-30
Patch: deberta success rate throttle

ryansingman:
See thread in backend PR

### DIFF
--- a/cleanlab_studio/internal/api/api.py
+++ b/cleanlab_studio/internal/api/api.py
@@ -652,7 +652,7 @@ async def tlm_prompt(
             await handle_tlm_client_error_from_resp(res, batch_index)
             await handle_tlm_api_error_from_resp(res, batch_index)
 
-            if not res_json.get("nli_deberta_success", True):
+            if not res_json.get("deberta_success", True):
                 raise TlmPartialSuccess("Partial failure on deberta call -- slowdown request rate.")
 
     finally:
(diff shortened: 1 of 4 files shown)


cleanlab_studio/internal/api/api.py
```diff
@@ -651,6 +652,9 @@ async def tlm_prompt(
             await handle_tlm_client_error_from_resp(res, batch_index)
             await handle_tlm_api_error_from_resp(res, batch_index)
 
+            if not res_json.get("nli_deberta_success", True):
```
huiwengoh, line 655: Isn't the key name `deberta_success` in the backend?

cleanlab_studio/internal/api/api.py
```diff
@@ -651,6 +652,9 @@ async def tlm_prompt(
...
+                raise TlmPartialSuccess("Partial failure on deberta call -- slowdown request rate.")
```
huiwengoh, line 656: Wouldn't this `raise` result in the `return` not actually returning the results to the user / trigger retries? Not entirely certain of the logic behind how the errors are handled here.

ryansingman: no, not currently -- the context manager in the rate handler will swallow this type of exception, making it so that execution continues (returning the result to users)

huiwengoh (approved):
lgtm!

PR #232, opened 2024-05-30
Improve data enrichment functions

huiwengoh:
- change how regex is being used: support both extraction (matching) and replacements
- set default model used as `claude-3-haiku`

blocked by the merge of https://github.com/cleanlab/cleanlab-studio/pull/230 (which allows the use to claude models)

### Sample usage

```python
data = pd.DataFrame({"inputs": ["1, 1, 1", "1, 2, 3", "1, 2, 3.5", "2, 4, 6"]})
prompt = "What is the sum of given numbers? Respond in the format 'The sum is: <sum>'. Numbers: {inputs}"

initial_results = enrich_data(studio, data, prompt, subset_indices=None)
initial_results = data.join(initial_results)

# try a bunch of prompts and regexes using process_regex (replacements)
# example 1: replace all non-numbers
process_regex(initial_results["metadata"], ('\D', '')) 

# example 2: (chaining regex replacements) remove the prefix The sum is:" and also any periods at the end of the sentence
# this sample is better (it does not incorrectly remove decimal points)
process_regex(initial_results["metadata"], [('.*The sum is:', ''), ('\.$', "")])

# example 3: simplify it even more by using matching
process_regex(initial_results["metadata"], ".*The sum is: (.*)")

# so now we can pass the regex directly in the enrich_data function
results = enrich_data(studio, data, prompt, regex=".*The sum is: (.*)", subset_indices=None)
results = data.join(results)
results
```

### Reviewer notes
@eytanhanig if you could verify that the expected use of regex is similar to what was scoped out for the backend API, that'd be great, thanks!

### DIFF
--- a/cleanlab_studio/internal/enrichment_utils.py
+++ b/cleanlab_studio/internal/enrichment_utils.py
@@ -9,11 +9,17 @@ from cleanlab_studio.errors import ValidationError
 from cleanlab_studio.studio.studio import Studio
 from cleanlab_studio.studio.trustworthy_language_model import TLMResponse
 
+Replacement = Tuple[Union[str, re.Pattern[str]], str]
+
 
 def get_prompt_outputs(
     studio: Studio, prompt: str, data: pd.DataFrame, **kwargs: Any
 ) -> List[Optional[TLMResponse]]:
     """Returns the outputs of the prompt for each row in the dataframe."""
+    default_tlm_options = {"model": "claude-3-haiku"}
+    tlm_options = kwargs.get("options", {})
+    kwargs["options"] = {**default_tlm_options, **tlm_options}
+
     tlm = studio.TLM(**kwargs)
     formatted_prompts = data.apply(lambda x: prompt.format(**x), axis=1).to_list()
     outputs = tlm.try_prompt(formatted_prompts)
@@ -37,41 +43,44 @@ def extract_df_subset(
     return subset_df
 
 
-def get_compiled_regex_list(
-    regex: Union[str, re.Pattern[str], List[re.Pattern[str]]]
-) -> List[re.Pattern[str]]:
-    """Compile the regex pattern(s) provided and return a list of compiled regex patterns."""
+def get_compiled_regex(regex: Union[str, re.Pattern[str]]) -> re.Pattern[str]:
+    """Compile the regex pattern(s) provided and return the compiled regex pattern."""
     if isinstance(regex, str):
-        return [re.compile(rf"{regex}")]
+        return re.compile(rf"{regex}")
     elif isinstance(regex, re.Pattern):
-        return [regex]
-    elif isinstance(regex, list):
         return regex
     else:
-        raise ValidationError(
-            "Passed in regex can only be type one of: str, re.Pattern, or list of re.Pattern."
-        )
+        raise ValidationError("Passed in regex can only be type one of: str, re.Pattern.")
 
 
-def get_regex_match(
-    response: str, regex_list: List[re.Pattern[str]], disable_warnings: bool
+def get_regex_replacement(
+    response: str, regex_replacements: Union[Replacement, List[Replacement]]
 ) -> Optional[str]:
-    """Extract the first match from the response using the provided regex patterns. Return first match if multiple exist.
-    Note: This function assumes the regex patterns each specify exactly 1 group that is the match group using ``'(<group>)'``."""
-    for regex_pattern in regex_list:
-        pattern_match = regex_pattern.match(response)
-        if pattern_match:
-            return pattern_match.group(1)
-    if not disable_warnings:
-        warnings.warn(
-            f"Your provided regex: {str(regex_list)} did not match a single LLM output across the dataset. This message is simply to inform you that the regex is not having any effect."
-        )
-    return None
+    """Performs regex replacements to the given string according to the given matching patterns and replacement strings."""
+    if isinstance(regex_replacements, tuple):
+        regex_replacements_list = [regex_replacements]
+    elif isinstance(regex_replacements, list):
+        regex_replacements_list = regex_replacements
+    else:
+        raise ValidationError("Passed in regex has to be either a tuple or a list of tuples.")
+
+    for replacement_pair in regex_replacements_list:
+        if not isinstance(replacement_pair, tuple) or len(replacement_pair) != 2:
+            raise ValidationError(
+                "Every item of the regex list must be a tuple that contains 2 items. TODO"
+            )
+
+        compiled_pattern = get_compiled_regex(replacement_pair[0])
+        replacement = replacement_pair[1]
+        response = compiled_pattern.sub(replacement, response)
+
+    return response
 
 
 def get_optimized_prompt(prompt: str, return_values: Optional[List[str]] = None) -> str:
     """Optimize the prompt by ammending original.
-    Adds a pre-prompt message if return_values are provided. This will help the LLM understand it's response must exactly match one of the return values."""
+    Adds a pre-prompt message if return_values are provided. This will help the LLM understand it's response must exactly match one of the return values.
+    """
 
     if return_values is not None:
         string_return_values = str(return_values).replace("'", "")
--- a/cleanlab_studio/internal/enrichment_utils.py
+++ b/cleanlab_studio/internal/enrichment_utils.py
@@ -9,7 +9,7 @@ from cleanlab_studio.errors import ValidationError
 from cleanlab_studio.studio.studio import Studio
 from cleanlab_studio.studio.trustworthy_language_model import TLMResponse
 
-Replacement = Tuple[Union[str, re.Pattern[str]], str]
+Replacement = Tuple[str, str]
 
 
 def get_prompt_outputs(
@@ -43,16 +43,6 @@ def extract_df_subset(
     return subset_df
 
 
-def get_compiled_regex(regex: Union[str, re.Pattern[str]]) -> re.Pattern[str]:
-    """Compile the regex pattern(s) provided and return the compiled regex pattern."""
-    if isinstance(regex, str):
-        return re.compile(rf"{regex}")
-    elif isinstance(regex, re.Pattern):
-        return regex
-    else:
-        raise ValidationError("Passed in regex can only be type one of: str, re.Pattern.")
-
-
 def get_regex_replacement(
     response: str, replacements: Union[Replacement, List[Replacement]]
 ) -> Optional[str]:
@@ -67,10 +57,11 @@ def get_regex_replacement(
     for replacement_pair in replacements_list:
         if not isinstance(replacement_pair, tuple) or len(replacement_pair) != 2:
             raise ValidationError(
-                "Every item of the regex list must be a tuple that contains 2 items. TODO"
+                "Every item of the regex list must be a tuple that contains 2 strings: "
+                "(the regex pattern to match, the string to replace the matched pattern with)"
             )
 
-        compiled_pattern = get_compiled_regex(replacement_pair[0])
+        compiled_pattern = re.compile(replacement_pair[0])
         replacement = replacement_pair[1]
         response = compiled_pattern.sub(replacement, response)
 
--- a/cleanlab_studio/internal/enrichment_utils.py
+++ b/cleanlab_studio/internal/enrichment_utils.py
@@ -54,17 +54,17 @@ def get_compiled_regex(regex: Union[str, re.Pattern[str]]) -> re.Pattern[str]:
 
 
 def get_regex_replacement(
-    response: str, regex_replacements: Union[Replacement, List[Replacement]]
+    response: str, replacements: Union[Replacement, List[Replacement]]
 ) -> Optional[str]:
     """Performs regex replacements to the given string according to the given matching patterns and replacement strings."""
-    if isinstance(regex_replacements, tuple):
-        regex_replacements_list = [regex_replacements]
-    elif isinstance(regex_replacements, list):
-        regex_replacements_list = regex_replacements
+    if isinstance(replacements, tuple):
+        replacements_list = [replacements]
+    elif isinstance(replacements, list):
+        replacements_list = replacements
     else:
         raise ValidationError("Passed in regex has to be either a tuple or a list of tuples.")
 
-    for replacement_pair in regex_replacements_list:
+    for replacement_pair in replacements_list:
         if not isinstance(replacement_pair, tuple) or len(replacement_pair) != 2:
             raise ValidationError(
                 "Every item of the regex list must be a tuple that contains 2 items. TODO"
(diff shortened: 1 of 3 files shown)


eytanhanig:
It's looking likely that we use something with a different syntax. Right now the strongest candidate looks like Google's re2. Here are the details:
* [Python package google-re2](https://pypi.org/project/google-re2/) - It includes a note about what differs from the standard `re` library, however I'm skeptical that's the only thing that's different.
* [Syntax for re2](https://github.com/google/re2/wiki/Syntax)

jwmueller:
I can't review this one (very hard to grok regex, needs playing with).

@eytanhanig we need to get this in asap, our deadline for delivering is next week. 

So please just review in the current form, we aren't going to be able to make this match the final API version anyway. It's ok if the regex syntax specifics are slightly different between this client-side version and the final API version.  We just want to ensure the code for this client-side version is correct, and are making small efforts to make it look closer to final API version (sounds like you haven't figured out the regex syntax 100% yet)

eytanhanig:
> It's ok if the regex syntax specifics are slightly different between this client-side version and the final API version

@jwmueller In that case I'm in total agreement that we should keep using `re` for the client-side implementation deadline.

cleanlab_studio/utils/data_enrichment/enrich.py
```diff
@@ -36,14 +36,17 @@ def enrich_data(
...
-            If a list of expressions is provided, the expressions are applied in order and first valid extraction is returned.
-
-            **Note:** Each regex pattern should specify 1 group that represents the desired characters to extract from the raw LLM response using parenthesis like so: ``'(<desired match group pattern>)'``.
-            **Example 1:** ``regex = r'.*The answer is: (Bird|[Rr]abbit).*'`` will extract strings that are the words 'Bird', 'Rabbit' or 'rabbit' after the characters "The answer is: " from the raw LLM response. This might be useful if your prompt instructed the LLM to respond in this format.
-            **Example 2:** ``regex = r'.*(\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Z|a-z]{2,}\b).*'`` will extract an email in the raw LLM response. Similar patterns can be used to extract a specifically structured part of the LLM response.
+        regex_replacements: A tuple or list of tuples each containing a pair of items:
```
eytanhanig, line 39: Some of the [existing Python Client documentation](https://github.com/cleanlab/cleanlab-studio/blob/10131ee806edfd642e753bf7a5536df213a4b64d/cleanlab_studio/studio/studio.py#L438) has the format of `verbose (bool, optional):`. Consider doing the same, especially if some of the args are optional.

cleanlab_studio/utils/data_enrichment/enrich.py
```diff
@@ -18,7 +18,7 @@ def enrich_data(
     prompt: str,
     data: pd.DataFrame,
     *,
-    regex: Optional[Union[str, re.Pattern[str], List[re.Pattern[str]]]] = None,
+    regex_replacements: Optional[Union[Replacement, List[Replacement]]] = None,
```
eytanhanig, line 21: I'm not sure how closely we want this to mirror what we currently have in `EnrichmentOptions`, however that's recently been switched to be `replacements`.

cleanlab_studio/utils/data_enrichment/enrich.py
```diff
@@ -33,81 +32,82 @@ def enrich_data(
...
+        subset_indices (Tuple[int, int] | List[int], optional): What subset of the supplied data rows to generate metadata for. If None, we run on all of the data.
             This can be either a list of unique indices or a range. These indices are passed into pandas ``.iloc`` method, so should be integers based on row order as opposed to row-index labels pointing to `df.index`.
             We advise against collecting results for all of your data at first. First collect results for a smaller data subset, and use this subset to experiment with different values of the `prompt` or `regex` arguments. Only once the results look good for your subset should you run on the full dataset.
-        metadata_column_name: Optional name for the returned metadata column. Name acts as a prefix appended to all additional columns that are returned.
-        disable_warnings: When True, warnings are disabled.
+        new_column_name (str): Optional name for the returned new metadata column. Name acts as a prefix appended to all additional columns that are returned.
```
eytanhanig, line 59: How about calling this the "enriched" column? Or just say "new column" or "column with the newly generated data".

eytanhanig (approved):

cleanlab_studio/utils/data_enrichment/enrich.py
```diff
@@ -15,14 +14,14 @@
...
+    prompt: str,
     *,
-    regex_replacements: Optional[Union[Replacement, List[Replacement]]] = None,
-    return_values: Optional[List[str]] = None,
+    replacements: Optional[Union[Replacement, List[Replacement]]] = None,
+    constrain_outputs: Optional[List[str]] = None,
     optimize_prompt: bool = True,
     subset_indices: Optional[Union[Tuple[int, int], List[int]]] = (0, 3),
-    metadata_column_name: str = "metadata",
+    new_column_name: str = "metadata",
```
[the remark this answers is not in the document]
huiwengoh: This is marked as optional in the API spec, will keep it optional here as well.
I can see how it could be nice (esp for new users randomly testing this out) to not have to worry about specifying a column name (less decisions lol)

eytanhanig (approved):

cleanlab_studio/utils/data_enrichment/enrich.py
```diff
@@ -33,112 +32,125 @@ def enrich_data(
...
-            **Example 2:** ``regex = r'.*(\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Z|a-z]{2,}\b).*'`` will extract an email in the raw LLM response. Similar patterns can be used to extract a specifically structured part of the LLM response.
-        return_values: List of all possible values for the `metadata` column.
+        regex (str | Replacement | List[Replacement], optional): A string, a tuple or list of tuples each containing:
+            - a string containing the regex pattern to match
+            - a string to replace the matched pattern with
+
+            If a string is passed in, a regex match will be performed and the matched pattern will be returned (if the pattern cannot be matched, it will return None).
+            If a tuple is passed in, the function will perform regex replacements on the response according to the given matching patterns and replacement strings.
+            If a list of tuples is passed in, the replacements are applied in the order they appear in the list.
+            Note that you cannot pass in a list of strings (chaining of multiple regex steps is only allowed for replacements).
```
jwmueller, line 46:
```suggestion
        regex (str | Replacement | List[Replacement], optional): A string, tuple, or list of tuples specifying regular expressions to apply for post-processing the raw LLM outputs.

            If a string value is passed in, a regex match will be performed and the matched pattern will be returned (if the pattern cannot be matched, None will be returned).
            Specifically the provided string will be passed into Python's `re.match()` method.
            Pass in a tuple `(R1, R2)` instead if you wish to perform find and replace operations rather than matching/extraction. 
            `R1` should be a string containing the regex pattern to match, and `R2` should be a string to replace matches with.
            Pass in a list of tuples instead if you wish to apply multiple replacements. Replacements will be applied in the order they appear in the list.
            Note that you cannot pass in a list of strings (chaining of multiple regex processing steps is only allowed for replacement operations).
```

cleanlab_studio/internal/enrichment_utils.py
```diff
@@ -9,11 +9,17 @@
 from cleanlab_studio.studio.studio import Studio
 from cleanlab_studio.studio.trustworthy_language_model import TLMResponse
 
+Replacement = Tuple[str, str]
```
ulya-tkch, line 12: This is intentionally changed from Union[re.Pattern[str], str] in case user wants to pass in compiled patterns, right?

cleanlab_studio/internal/enrichment_utils.py
```diff
@@ -37,83 +43,97 @@ def extract_df_subset(
...
+    if isinstance(regex, tuple):
+        regex_list = [regex]
     elif isinstance(regex, list):
-        return regex
+        regex_list = regex
     else:
         raise ValidationError(
-            "Passed in regex can only be type one of: str, re.Pattern, or list of re.Pattern."
+            "Passed in regex has to be either a string (pattern to match), a tuple (pair of patterns to match and replace), "
+            "or a list of tuples (pairs of patterns to match and replace)."
```
ulya-tkch, line 68:
```suggestion
           ", a list of tuples (pairs of patterns to match and replace) or a list of strings ."
```

cleanlab_studio/internal/enrichment_utils.py
```diff
@@ -37,83 +43,97 @@ def extract_df_subset(
     return subset_df
 
 
-def get_compiled_regex_list(
-    regex: Union[str, re.Pattern[str], List[re.Pattern[str]]]
-) -> List[re.Pattern[str]]:
-    """Compile the regex pattern(s) provided and return a list of compiled regex patterns."""
+def get_regex_match_or_replacement(
```
ulya-tkch, line 46: Just confirming, only a single pattern is used for matching (re.search) but a list of patterns can be used for replacements (re.sub). Previously, this was build that a list of patterns could be used for matching

cleanlab_studio/internal/enrichment_utils.py
```diff
@@ -37,83 +43,97 @@ def extract_df_subset(
...
-            f"Your provided regex: {str(regex_list)} did not match a single LLM output across the dataset. This message is simply to inform you that the regex is not having any effect."
-        )
-    return None
+        if not isinstance(replacement_pair, tuple) or len(replacement_pair) != 2:
+            raise ValidationError(
+                "Every item of the regex list must be a tuple that contains 2 strings: "
+                "(the regex pattern to match, the string to replace the matched pattern with)"
+            )
 
+        compiled_pattern = re.compile(replacement_pair[0])
```
ulya-tkch, line 81:
```suggestion
        # check regex is well formatted
        compiled_pattern = re.compile(replacement_pair[0])
```
Below is the function I used, no worries adding it now but I though i'd share. It would have to be modified a bit for this code
```python3
def check_regex_well_formed(
    regex_list: List[Tuple[re.Pattern[str], str]]
) -> None:
    """Function checks the regex for any too complex patterns that would take unexpectedly long to compute"""
    complex_patterns = [
        (re.compile(r'\\\d'), "backreference"), # backreference pattern
        (re.compile(r'\(\?<=|\(\?<!|\(\?=|\(\?!'), "lookahead/lookbehind"), # lookahead and lookbehind patterns
    ]

    for regex_pattern, sub_text in regex_list:
        for complex_pattern, complex_pattern_name in complex_patterns:
            if complex_pattern.search(regex_pattern):
                raise ValueError(f"Regular expressions can not contain {complex_pattern_name}. Only simplified regular expressions are allowed." )
            if complex_pattern.search(re.compile(sub_text)):   # if regex is applied in-order then this is necessary
                raise ValueError(f"Substitute text for regular expressions can not contain {complex_pattern_name}. Only simplified text is allowed." )
```

cleanlab_studio/utils/data_enrichment/enrich.py
```diff
@@ -33,112 +32,125 @@ def enrich_data(
...
-            If a list of expressions is provided, the expressions are applied in order and first valid extraction is returned.
-
-            **Note:** Each regex pattern should specify 1 group that represents the desired characters to extract from the raw LLM response using parenthesis like so: ``'(<desired match group pattern>)'``.
-            **Example 1:** ``regex = r'.*The answer is: (Bird|[Rr]abbit).*'`` will extract strings that are the words 'Bird', 'Rabbit' or 'rabbit' after the characters "The answer is: " from the raw LLM response. This might be useful if your prompt instructed the LLM to respond in this format.
-            **Example 2:** ``regex = r'.*(\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Z|a-z]{2,}\b).*'`` will extract an email in the raw LLM response. Similar patterns can be used to extract a specifically structured part of the LLM response.
-        return_values: List of all possible values for the `metadata` column.
+        regex (str | Replacement | List[Replacement], optional): A string, tuple, or list of tuples specifying regular expressions to apply for post-processing the raw LLM outputs.
```
ulya-tkch, line 39: I would mention regex is applied and results can be seen in {new_column_name} but the  {new_column_name}log column contains raw outputs pre-regex.

cleanlab_studio/utils/data_enrichment/enrich.py
```diff
@@ -33,112 +32,125 @@ def enrich_data(
...
+
+            **Example 1:** ``regex = '.*The answer is: (Bird|[Rr]abbit).*'`` will extract strings that are the words 'Bird', 'Rabbit' or 'rabbit' after the characters "The answer is: " from the raw response.
+            **Example 2:** ``regex = [('True', 'T'), ('False', 'F')]`` will replace the words True and False with T and F.
+            **Example 3:** ``regex = (' Explanation:.*', '') will remove everything after and including the words "Explanation:".
+            For instance, the response "True. Explanation: 3+4=7, and 7 is an odd number." would return "True" after the regex replacement.
+        constrain_outputs (List[str], optional): List of all possible output values for the `metadata` column.
             If specified, every entry in the `metadata` column will exactly match one of these values (for less open-ended data enrichment tasks). If None, the `metadata` column can contain arbitrary values (for more open-ended data enrichment tasks).
             After your regex is applied, there may be additional transformations applied to ensure the returned value is one of these.
```
ulya-tkch, line 58:
```suggestion
           There may be additional transformations applied to ensure the returned value is one of these. If regex is also specified, then these transformations occur after your regex is applied.
```

cleanlab_studio/utils/data_enrichment/enrich.py
```diff
@@ -33,112 +32,125 @@ def enrich_data(
...
+        A DataFrame that contains `metadata` and `trustworthiness` columns related to the prompt in order of original data. Some columns names will have `new_column_name` prepended to them.
+        `metadata` column = responses to the prompt and other data mutations if `replacements` or `constrain_outputs` is not specified.
         `trustworthiness` column = trustworthiness of the prompt responses (which ignore the data mutations).
-        **Note**: If you specified the `regex` or `return_values` arguments, some additional transformations may be applied to raw LLM outputs to produce the returned values. In these cases, an additional `log` column will be added to the returned DataFrame that records the raw LLM outputs (feel free to disregard these).
+        **Note**: If you specified the `replacements` or `constrain_outputs` arguments, some additional transformations may be applied to raw LLM outputs to produce the returned values. In these cases, an additional `log` column will be added to the returned DataFrame that records the raw LLM outputs (feel free to disregard these).
```
ulya-tkch, line 72: No argument called `replacements`, dod you mean `regex`

cleanlab_studio/internal/enrichment_utils.py
huiwengoh: Yes, we are no longer allowing compile regex patterns (only strings that we will compile at runtime)

cleanlab_studio/utils/data_enrichment/enrich.py
```diff
@@ -33,112 +32,125 @@ def enrich_data(
...
 
-    **Note:** Regex patterns should each specify exactly 1 group that is represents the desired characters to be extracted from the raw response using parenthesis like so ``'(<desired match group pattern>)'``.
-    **Example 1:** `r'.*The answer is: (Bird|[Rr]abbit).*'` will extract strings that are the words 'Bird', 'Rabbit' or 'rabbit' after the characters "The answer is: " from the raw response text.
-    **Example 2:** `r'.*(\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Z|a-z]{2,}\b).*'` will match an email in the raw response LLM response.
+    **Example 1:** ``regex = '.*The answer is: (Bird|[Rr]abbit).*'`` will extract strings that are the words 'Bird', 'Rabbit' or 'rabbit' after the characters "The answer is: " from the raw response.
+    **Example 2:** ``regex = [('True', 'T'), ('False', 'F')]`` will replace the words True and False with T and F.
+    **Example 3:** ``regex = (' Explanation:.*', '') will remove everything after and including the words "Explanation:".
+    For instance, the response "True. Explanation: 3+4=7, and 7 is an odd number." would return "True" after the regex replacement.
```
ulya-tkch, line 142: Wouldn't this return "True."

cleanlab_studio/utils/data_enrichment/enrich.py
```diff
@@ -33,112 +32,125 @@ def enrich_data(
...
+            Note that you cannot pass in a list of strings (chaining of multiple regex processing steps is only allowed for replacement operations).
+
+            These tuples specify the desired patterns to match and replace from the raw LLM response,
+            This regex processing is useful in settings where you are unable to prompt the LLM to generate valid outputs 100% of the time,
+            but can easily transform the raw LLM outputs to be valid through regular expressions that extract and replace parts of the raw output string.
+
+            **Example 1:** ``regex = '.*The answer is: (Bird|[Rr]abbit).*'`` will extract strings that are the words 'Bird', 'Rabbit' or 'rabbit' after the characters "The answer is: " from the raw response.
+            **Example 2:** ``regex = [('True', 'T'), ('False', 'F')]`` will replace the words True and False with T and F.
+            **Example 3:** ``regex = (' Explanation:.*', '') will remove everything after and including the words "Explanation:".
+            For instance, the response "True. Explanation: 3+4=7, and 7 is an odd number." would return "True" after the regex replacement.
```
ulya-tkch, line 55: Wouldn't this return "True."

ulya-tkch (approved):
Left few small comments, code LGTM!

cleanlab_studio/internal/enrichment_utils.py
huiwengoh: That is correct, the options now are:
- a string: matching/extraction (only one string pattern is allowed)
- a tuple: replacement
- a list of tuples: a series of replacements sequentially

cleanlab_studio/utils/data_enrichment/enrich.py
huiwengoh: Ah sorry yes this was changed multiple things across the past few days, will update

cleanlab_studio/utils/data_enrichment/enrich.py
huiwengoh: > **Example 3:** ``regex = (' Explanation:.*', '') will remove everything after and including the words "Explanation:".
>     For instance, the response "True. Explanation: 3+4=7, and 7 is an odd number." would return "True" after the regex replacement.

This is stated here? Not sure if something was misunderstood

huiwengoh: Oh nvm I realized it was missing the period.

cleanlab_studio/utils/data_enrichment/enrich.py
ulya-tkch: The tutorial notebook uses the variable "replacements" in a lot of places. I would standardize calling it regex if you can

cleanlab_studio/utils/data_enrichment/enrich.py
ulya-tkch: Yep I think that regex would still include "."

PR #233, opened 2024-06-03
Add ability to add, delete, and get enrichment project(s)

eytanhanig:
This adds the ability to add, delete, and list Enrichment Projects.

#### Tests:
```py
>>> ep = studio.create_enrichment_project(dataset_id="ad409cee4080496587e770c9c6ba97b2", name="hello-world")

>>> ep.to_dict()
{'id': 'd498d2e62f9a4ec6910ff45cd09e47d2', 'name': 'hello-world', 'created_at': datetime.datetime(2024, 6, 6, 0, 1, 55), 'updated_at': datetime.datetime(2024, 6, 6, 0, 1, 55)}

>>> studio.get_enrichment_projects()
[<cleanlab_studio.studio.enrichment.EnrichmentProject object at 0x10041d750>, <cleanlab_studio.studio.enrichment.EnrichmentProject object at 0x12aeb4850>, <cleanlab_studio.studio.enrichment.EnrichmentProject object at 0x12aeb4ee0>, <cleanlab_studio.studio.enrichment.EnrichmentProject object at 0x12aeb4f10>]

>>> studio.delete_enrichment_project(project_id=ep.id)
Successfully deleted enrichment project: d498d2e62f9a4ec6910ff45cd09e47d2
```

### DIFF
--- a/cleanlab_studio/internal/api/api.py
+++ b/cleanlab_studio/internal/api/api.py
@@ -49,6 +49,7 @@ base_url = os.environ.get("CLEANLAB_API_BASE_URL", "https://api.cleanlab.ai/api"
 cli_base_url = f"{base_url}/cli/v0"
 upload_base_url = f"{base_url}/upload/v1"
 dataset_base_url = f"{base_url}/datasets"
+enrichment_base_url = f"{base_url}/enrichment/v0"
 project_base_url = f"{base_url}/projects"
 cleanset_base_url = f"{base_url}/cleansets"
 model_base_url = f"{base_url}/v1/deployment"
@@ -572,6 +573,72 @@ def get_deployed_model_info(api_key: str, model_id: str) -> Dict[str, str]:
     handle_api_error(res)
 
     return cast(Dict[str, str], res.json())
+
+
+def create_enrichment_project(
+    api_key: str,
+    dataset_id: str,
+    name: str,
+    target_column_in_dataset: str,
+) -> JSONDict:
+    """Create a new enrichment project."""
+    check_uuid_well_formed(dataset_id, "dataset ID")
+    request_json = dict(
+        dataset_id=dataset_id,
+        name=name,
+        target_column_in_dataset=target_column_in_dataset,
+    )
+    res = requests.post(
+        f"{enrichment_base_url}/projects",
+        headers=_construct_headers(api_key),
+        json=request_json,
+    )
+    handle_api_error(res)
+    res_json: JSONDict = res.json()
+    return cast(JSONDict, res_json)
+
+
+def delete_enrichment_project(api_key: str, project_id: str) -> None:
+    """Delete an enrichment project."""
+    check_uuid_well_formed(project_id, "Enrichment Project ID")
+    res = requests.delete(
+        f"{enrichment_base_url}/projects/{project_id}", headers=_construct_headers(api_key)
+    )
+    handle_api_error(res)
+    return
+
+
+def get_enrichment_project(api_key: str, project_id: str) -> JSONDict:
+    """Get an existing enrichment project."""
+    check_uuid_well_formed(project_id, "Enrichment Project ID")
+    res = requests.get(
+        f"{enrichment_base_url}/projects/{project_id}",
+        headers=_construct_headers(api_key),
+    )
+    handle_api_error(res)
+    res_json: JSONDict = res.json()
+    return cast(JSONDict, res_json)
+
+
+def get_enrichment_projects(api_key: str) -> List[JSONDict]:
+    """Get a list of all enrichment projects."""
+    all_projects: List[JSONDict] = []
+    page: Optional[int] = 1
+
+    while page is not None:
+        res = requests.get(
+            f"{enrichment_base_url}/projects?page={page}",
+            headers=_construct_headers(api_key),
+        )
+        handle_api_error(res)
+        res_json: JSONDict = res.json()
+
+        if "projects" in res_json:
+            all_projects.extend(res_json["projects"])
+
+        page = res_json.get("next_page", None)
+
+    return all_projects
 
 
 def tlm_retry(func: Callable[..., Any]) -> Callable[..., Any]:
--- a/cleanlab_studio/internal/api/api.py
+++ b/cleanlab_studio/internal/api/api.py
@@ -604,7 +604,6 @@ def delete_enrichment_project(api_key: str, project_id: str) -> None:
         f"{enrichment_base_url}/projects/{project_id}", headers=_construct_headers(api_key)
     )
     handle_api_error(res)
-    return
 
 
 def get_enrichment_project(api_key: str, project_id: str) -> JSONDict:
@@ -618,7 +617,7 @@ def get_enrichment_project(api_key: str, project_id: str) -> JSONDict:
     return cast(JSONDict, res.json())
 
 
-def get_enrichment_projects(api_key: str) -> List[JSONDict]:
+def list_all_enrichment_projects(api_key: str) -> List[JSONDict]:
     """Get a list of all enrichment projects."""
     all_projects: List[JSONDict] = []
     page: Optional[int] = 1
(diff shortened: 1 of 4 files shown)


cleanlab_studio/studio/enrichment.py
```diff
@@ -0,0 +1,67 @@
...
+        self,
+        api_key: str,
+        id: str,
+        name: str,
+        target_column_in_dataset: str,
+        created_at_string: Optional[str] = None,
+    ) -> None:
+        """Initialize an EnrichmentProject.
+
+        **Objects of this class are not meant to be constructed directly.** Instead, use [`Studio.get_enrichment_project()`](../studio/#method-get_enrichment_project).
```
novking, line 38: it can have `to_json` and `from_json`. to the user will have a true project object to create with.

cleanlab_studio/internal/api/api.py
```diff
@@ -574,6 +575,70 @@ def get_deployed_model_info(api_key: str, model_id: str) -> Dict[str, str]:
...
+    check_uuid_well_formed(project_id, "Enrichment Project ID")
+    res = requests.get(
+        f"{enrichment_base_url}/projects/{project_id}",
+        headers=_construct_headers(api_key),
+    )
+    handle_api_error(res)
+    return cast(JSONDict, res.json())
+
+
+def get_enrichment_projects(api_key: str) -> List[JSONDict]:
```
novking, line 621: as low-level api, it doesn't need to fetch all projects. but provide a way to do that. 
i.e. `def list_enrichment_project(api_key: str, next_page: int = 1)`

cleanlab_studio/studio/studio.py
```diff
@@ -396,6 +397,99 @@ def download_embeddings(
...
+            for the new enrichment project.
+        """
+        enrichment_project_dict = api.create_enrichment_project(
+            api_key=self._api_key,
+            dataset_id=dataset_id,
+            name=name,
+            target_column_in_dataset=target_column_in_dataset,
+        )
+
+        return enrichment.EnrichmentProject(
```
novking, line 425: this is what I mean. 
instead of manually put the dict value one by one. 
do `EnrichmentProject.from_dict(enrichment_project_dict)` hide everything into the object definition.

novking (approved):
No blockers, all the comments are suggestions to be more OOP.

novking:
Another suggestion. Put your tests in the PR description.

cleanlab_studio/internal/api/api.py
eytanhanig: How about changing this to `list_all_enrichment_projects()` and adding a pagination API if we need it later?

cleanlab_studio/studio/enrichment.py
eytanhanig: 1. Because this requires `api_key` I'm not sure we want to expose `from_json()` to the end user.
2. `to_json()` has been added.

cleanlab_studio/studio/studio.py
```diff
@@ -396,6 +397,99 @@ def download_embeddings(
...
+        )
+
+    def delete_enrichment_project(self, project_id: str) -> None:
+        """
+        Deletes an Enrichment Project from Cleanlab Studio.
+
+        Args:
+            project_id: ID of enrichment project to delete.
+        """
+        api.delete_enrichment_project(self._api_key, project_id=project_id)
```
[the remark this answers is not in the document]
eytanhanig: I believe that `handle_api_error(res)` in the API takes care of this for us.

cleanlab_studio/studio/studio.py
eytanhanig: Unfortunately `created_at_string` complicates this.

cleanlab_studio/internal/api/api.py
novking: OK. 
This is not the best practice, just fyi

cleanlab_studio/studio/enrichment.py
novking: so the problem remains, user will use a dict instead of `EnrichmentProject`. 
is this a comment pattern in studio to require api_key explicitly. for cli like, aws, docker, they just read from a local file or ENV VAR. 
I would suggest to think about if we can do that.

cleanlab_studio/studio/studio.py
novking: I'm not sure I understand why created_at makes things complicate.
it's a value from backend with fixed format, `from_dict`or `from_json` can just adapt the format. no?

novking (approved):
I would suggest for `Test` in the description. put the test value/output.
It's the PR owner's responsibility to prove the PR works as expected. the evident should be the results from `Test` section.

cleanlab_studio/internal/api/api.py
eytanhanig: >  This is not the best practice, just fyi

What do you think we should return to the user?

eytanhanig:
> so the problem remains, user will use a dict instead of `EnrichmentProject`. 
> is this a comment pattern in studio to require api_key explicitly. for cli like, aws, docker, they just read from a local file or ENV VAR. 
> I would suggest to think about if we can do that. 

@novking Yes, this is the pattern that we use. I agree that we should change it. Note that there's support for switching to a more object-oriented structure for this library in the future.

cleanlab_studio/studio/studio.py
eytanhanig: I've updated `EnrichmentProject`s `created_at` arg to take string as well, and now do the conversion inside the class.

cleanlab_studio/studio/studio.py
```diff
@@ -396,6 +397,93 @@ def download_embeddings(
...
+            api_key=self._api_key,
+            project_id=project_id,
+        )
+
+        return enrichment.EnrichmentProject(
+            self._api_key,
+            id=project_id,
+            name=enrichment_project_dict["name"],
+            created_at=enrichment_project_dict["created_at"],
+        )
```
novking, line 463: + `updated_at`

cleanlab_studio/studio/studio.py
```diff
@@ -396,6 +397,93 @@ def download_embeddings(
...
+        enrichment_project_dicts = api.list_all_enrichment_projects(
+            api_key=self._api_key,
+        )
+
+        return [
+            enrichment.EnrichmentProject(
+                self._api_key,
+                id=enrichment_project_dict["id"],
+                name=enrichment_project_dict["name"],
+            )
```
novking, line 483: i think add timestamp is better. user can sort in their own way and find things they need. 
suggest + `created_at` and `updated_at`. to make it consistent with other methods return.

cleanlab_studio/studio/studio.py
```diff
@@ -396,6 +397,93 @@ def download_embeddings(
...
+            name=name,
+            dataset_id=dataset_id,
+        )
+
+        return enrichment.EnrichmentProject(
+            api_key=self._api_key,
+            id=enrichment_project_dict["id"],
+            name=enrichment_project_dict["name"],
+            created_at=enrichment_project_dict["created_at"],
+        )
```
novking, line 427: + `updated_at`. i know the update time is not necessary. but we should keep all four fields across all methods. i.e., id, name, created_at, updated_at.

novking (approved):
suggestions are minors. approved.

cleanlab_studio/studio/enrichment.py
eytanhanig: I strongly agree that this pattern is not ideal and would like to see it tackled in a major release increment.

cleanlab_studio/studio/studio.py
eytanhanig: I decided to not store `updated_at` so that whenever a user attempts to look up this field they get the correct value.

cleanlab_studio/studio/studio.py
eytanhanig: I've added `created_at`.

PR #235, opened 2024-06-06
change abort() to AuthError

kat-wicks:
Use AuthError instead of abort() for Invalid API keys to reduce the alert noise from wrong keys


axl1313 (approved):

axl1313:
Can you fix CI and format before merging

PR #238, opened 2024-06-11
Add logs to TLM outputs + allow perplexity as input

huiwengoh:
- Adds a new TLMOption, `log`, which allows users to specify what additional logs/metadata they want returned.
  If logs is specified (by default it will be `None`):
  - `tlm.prompt()` will return the same dictionary with an additional `log` key-value pair
  - `tlm.get_trustworthiness_score` will return a dictionary (with `trustworthiness_score` and `log` as keys) instead of float

- Adds an new optional keyword argument to `tlm.get_trustworthiness_score` named `perplexity` which allows users to pass in perplexity scores (from their own LLM, or anywhere else) to consider as part of the scoring computation

### Sample workflow:

```python
tlm_prompt = studio.TLM("base", options={ "log": ["perplexity"]})
res = tlm_prompt.prompt(prompt)
res 

>> {"response": ..., trustworthiness_score": ..., "log": {"perplexity": 0.5}}
```

We see that the results above has an extra key `log` which gives us the metadata, that perplexity can now be used in the `get_trustworthiness_score` method as such:

```python
tlm_score = studio.TLM()
tlm_score.get_trustworthiness_score(prompt, res["response"], perplexity=res["log"]["perplexity"])
```

Sample batch prompt usage:
```python
prompt = ["What is 1+1?", "What is the colour of water?"]

tlm_prompt = studio.TLM("base", options={"log": ["perplexity"]})
res = tlm_prompt.prompt(prompt)
res

tlm_score = studio.TLM()
tlm_score.get_trustworthiness_score(prompt, [r["response"] for r in res], perplexity=[r["log"]["perplexity"] for r in res])
```

### DIFF
--- a/cleanlab_studio/internal/api/api.py
+++ b/cleanlab_studio/internal/api/api.py
@@ -667,7 +667,6 @@ async def tlm_get_confidence_score(
     api_key: str,
     prompt: str,
     response: str,
-    input_metadata: Optional[JSONDict],
     quality_preset: str,
     options: Optional[JSONDict],
     rate_handler: TlmRateHandler,
@@ -681,7 +680,6 @@ async def tlm_get_confidence_score(
         api_key (str): studio API key for auth
         prompt (str): prompt for TLM to get confidence score for
         response (str): response for TLM to get confidence score for
-        input_metadata (JSONDict): additional input metadata for TLM
         quality_preset (str): quality preset to use to generate confidence score
         options (JSONDict): additional parameters for TLM
         rate_handler (TlmRateHandler): concurrency handler used to manage TLM request rate
@@ -703,7 +701,6 @@ async def tlm_get_confidence_score(
                 json=dict(
                     prompt=prompt,
                     response=response,
-                    input_metadata=input_metadata or {},
                     quality=quality_preset,
                     options=options or {},
                 ),
--- a/cleanlab_studio/internal/api/api.py
+++ b/cleanlab_studio/internal/api/api.py
@@ -667,6 +667,7 @@ async def tlm_get_confidence_score(
     api_key: str,
     prompt: str,
     response: str,
+    logprobs: float,
     quality_preset: str,
     options: Optional[JSONDict],
     rate_handler: TlmRateHandler,
@@ -680,6 +681,7 @@ async def tlm_get_confidence_score(
         api_key (str): studio API key for auth
         prompt (str): prompt for TLM to get confidence score for
         response (str): response for TLM to get confidence score for
+        logprobs (float): log probabilities associated with the given responses
         quality_preset (str): quality preset to use to generate confidence score
         options (JSONDict): additional parameters for TLM
         rate_handler (TlmRateHandler): concurrency handler used to manage TLM request rate
@@ -699,7 +701,11 @@ async def tlm_get_confidence_score(
             res = await client_session.post(
                 f"{tlm_base_url}/get_confidence_score",
                 json=dict(
-                    prompt=prompt, response=response, quality=quality_preset, options=options or {}
+                    prompt=prompt,
+                    response=response,
+                    logprobs=logprobs,
+                    quality=quality_preset,
+                    options=options or {},
                 ),
                 headers=_construct_headers(api_key),
             )
--- a/cleanlab_studio/internal/api/api.py
+++ b/cleanlab_studio/internal/api/api.py
@@ -667,7 +667,7 @@ async def tlm_get_confidence_score(
     api_key: str,
     prompt: str,
     response: str,
-    logprobs: float,
+    input_metadata: Optional[JSONDict],
     quality_preset: str,
     options: Optional[JSONDict],
     rate_handler: TlmRateHandler,
@@ -681,7 +681,7 @@ async def tlm_get_confidence_score(
         api_key (str): studio API key for auth
         prompt (str): prompt for TLM to get confidence score for
         response (str): response for TLM to get confidence score for
-        logprobs (float): log probabilities associated with the given responses
+        input_metadata (JSONDict): additional input metadata for TLM
         quality_preset (str): quality preset to use to generate confidence score
         options (JSONDict): additional parameters for TLM
         rate_handler (TlmRateHandler): concurrency handler used to manage TLM request rate
@@ -703,7 +703,7 @@ async def tlm_get_confidence_score(
                 json=dict(
                     prompt=prompt,
                     response=response,
-                    logprobs=logprobs,
+                    input_metadata=input_metadata or {},
                     quality=quality_preset,
                     options=options or {},
                 ),
--- a/cleanlab_studio/internal/tlm/validation.py
+++ b/cleanlab_studio/internal/tlm/validation.py
@@ -194,8 +194,9 @@ def validate_tlm_options(options: Any) -> None:
                 )
 
 
-def process_get_trustworthiness_score_kwargs(
-    prompt: Union[str, Sequence[str]], kwargs_dict: Dict[str, Any]
+def process_response_and_kwargs(
+    response: Union[str, Sequence[str]],
+    kwargs_dict: Dict[str, Any],
 ) -> Union[Dict[str, Any], List[Dict[str, Any]]]:
 
     invalid_kwargs = set(kwargs_dict.keys()) - TLM_VALID_GET_TRUSTWORTHINESS_SCORE_KWARGS
@@ -207,21 +208,21 @@ def process_get_trustworthiness_score_kwargs(
     # checking validity/format of each input kwarg, each one might require a different format
     for key, val in kwargs_dict.items():
         if key == "perplexity":
-            if isinstance(prompt, str):
+            if isinstance(response, str):
                 if not (val is None or isinstance(val, float) or isinstance(val, int)):
                     raise ValidationError(
-                        f"Invalid type {type(val)}, perplexity should be a float when prompt is a str."
+                        f"Invalid type {type(val)}, perplexity should be a float when response is a str."
                     )
                 if val is not None and not 0 <= val <= 1:
                     raise ValidationError("Perplexity values must be between 0 and 1")
 
-            elif isinstance(prompt, Sequence):
+            elif isinstance(response, Sequence):
                 if not isinstance(val, Sequence):
                     raise ValidationError(
-                        f"Invalid type {type(val)}, perplexity should be a sequence when prompt is a sequence"
+                        f"Invalid type {type(val)}, perplexity should be a sequence when response is a sequence"
                     )
-                if len(prompt) != len(val):
-                    raise ValidationError("Length of the prompt and perplexity lists must match.")
+                if len(response) != len(val):
+                    raise ValidationError("Length of the response and perplexity lists must match.")
                 if not all(v is None or 0 <= v <= 1 for v in val):
                     raise ValidationError("Perplexity values must be between 0 and 1")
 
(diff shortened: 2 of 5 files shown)


cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -371,22 +376,29 @@ async def _prompt_async(
...
+            tlm_response["logs"] = response_json["logs"]
+
+        return cast(TLMResponse, tlm_response)
+
     def get_trustworthiness_score(
         self,
         prompt: Union[str, Sequence[str]],
         response: Union[str, Sequence[str]],
-    ) -> Union[float, List[float]]:
+        logprobs: Optional[Union[float, Sequence[float]]] = None,
```
jwmueller, line 393:
```suggestion
        *,
        logprobs: Optional[Union[float, Sequence[float]]] = None,
```

jwmueller: I think this should be prob, not probs (since it's just one number) 
or could consider argname `perplexity`

jwmueller: make sure to apply these ^ changes everywhere

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -371,22 +376,29 @@ async def _prompt_async(
...
+    ) -> Union[TLMScoreResponse, List[TLMScoreResponse]]:
         """Computes trustworthiness score for arbitrary given prompt-response pairs.
 
         Args:
             prompt (str | Sequence[str]): prompt (or list of prompts) for the TLM to evaluate
             response (str | Sequence[str]): existing response (or list of responses) associated with the input prompts.
                 These can be from any LLM or human-written responses.
+            logprobs (float | Sequence[float]): log probabilities (or list of log probs) associated with the given responses.
```
jwmueller, line 401:
```suggestion
            logprob (float | Sequence[float]): Optional log probability (or list of logprobs) associated with the given response(s).
```

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -607,13 +646,16 @@ class TLMOptions(TypedDict):
         This self-reflection forms a big part of the trustworthiness score, helping quantify aleatoric uncertainty associated with challenging prompts
         and helping catch answers that are obviously incorrect/bad for a prompt asking for a well-defined answer that LLMs should be able to handle.
         Setting this to False disables the use of self-reflection and may produce worse TLM trustworthiness scores, but will reduce costs/runtimes.
+
+        logs (List[str], default = None): specify what additional logs or metadata should be returned. Types of logs available: ["logprobs"].
```
jwmueller, line 650:
```suggestion
        logs (List[str], default = None): Optionally specify additional logs or metadata to return.
```

jwmueller: note I suggest to all this `log` instead (as a verb)

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -552,10 +588,13 @@ class TLMResponse(TypedDict):
         trustworthiness_score (float, optional): score between 0-1 corresponding to the trustworthiness of the response.
         A higher score indicates a higher confidence that the response is correct/trustworthy. The trustworthiness score
         is omitted if TLM is run with quality preset "base".
+
+        logs (dict, optional): a dictionary containing additional logs and metadata from the LLM call.
```
jwmueller, line 592:
```suggestion
        log (dict, optional): a dictionary containing additional logs and metadata returned from the LLM call only if the `log` key was specified in TLMOptions.
```

jwmueller: how come your new type `TLMScoreResponse` is defined very differently to this `TLMResponse` object (in a different file, not as typedict, etc)?  

Couldn't `TLMScoreResponse` be defined here with a simple doctring: `Similar dict to TLMResponse, except without the response key.`

cleanlab_studio/studio/trustworthy_language_model.py
huiwengoh: Will change everything to `perplexity` (including the key name in the `log` dictionary).
Reason being that log probabilities are not on the 0-1 scale and are much less readable, I think it makes more sense to have something on the 0-1 scale (hence perplexity). Lmk if you have opinions otherwise.

cleanlab_studio/studio/trustworthy_language_model.py
huiwengoh: I agree this is a little confusing.
The difference is that `TLMResponse` defines a typed dict - the definition specifies what keys/typing the dict will have. Whereas `TLMScoreResponse` is just a type alias - because the `get_trustworthiness_score` method can either return a float or a dictionary, hence the type alias:
```
TLMScoreResponse = Union[float, Dict[str, Any]]
```

> Couldn't TLMScoreResponse be defined here with a simple doctring: Similar dict to TLMResponse, except without the response key.

The reason why this would not work is because the `get_trustworthiness_score` by default returns just a float (not a dict like `TLMResponse`. 

I think renaming the `TLMScoreResponse` type alias would be the best way to keep this the least confusing.

cleanlab_studio/studio/trustworthy_language_model.py
jwmueller: great but let's have this go into get_trustworthiness_score via kwargs only, so we dont need any documentation (they argname in kwargs can be perplexity)

cleanlab_studio/internal/tlm/validation.py
```diff
@@ -178,3 +180,69 @@ def validate_tlm_options(options: Any) -> None:
...
+            f"Invalid kwargs provided: {invalid_kwargs}. Valid kwargs include: {TLM_VALID_LOG_OPTIONS}"
+        )
+
+    # checking validity/format of each input kwarg, each one might require a different format
+    for key, val in kwargs_dict.items():
+        if key == "perplexity":
+            if val is None or isinstance(val, float) or isinstance(val, int):
+                if not isinstance(prompt, str):
+                    raise ValidationError(
+                        f"Invalid type {type(val)}, perplexity should be a float if prompt is a str, and should be a sequence if prompt is a sequence"
```
ulya-tkch, line 212: This should be relevant to prompt not being a string. Error message wrong

cleanlab_studio/internal/tlm/validation.py
```diff
@@ -178,3 +180,69 @@ def validate_tlm_options(options: Any) -> None:
...
+                    )
+                # TODO: should we raise warning if perplexity is None?
+                if val is not None and not 0 <= val <= 1:
+                    raise ValidationError("Perplexity values must be between 0 and 1")
+
+            elif isinstance(val, Sequence):
+                if not isinstance(prompt, Sequence) or isinstance(prompt, str):  # str is a Sequence
+                    raise ValidationError(
+                        f"Invalid type {type(val)}, perplexity should be a float if prompt is a str, and should be a sequence if prompt is a sequence"
```
ulya-tkch, line 221: This should be relevant to prompt not being a string/Sequence. Error message wrong

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -61,8 +62,11 @@ def __init__(
                 f"Invalid quality preset {quality_preset} -- must be one of {_VALID_TLM_QUALITY_PRESETS}"
             )
 
+        self._return_log = False
         if options is not None:
             validate_tlm_options(options)
+            if "log" in options.keys():  # TODO: figure out check
```
ulya-tkch, line 68: What is this TODO?

cleanlab_studio/internal/api/api.py
```diff
@@ -699,7 +701,11 @@ async def tlm_get_confidence_score(
             res = await client_session.post(
                 f"{tlm_base_url}/get_confidence_score",
                 json=dict(
-                    prompt=prompt, response=response, quality=quality_preset, options=options or {}
+                    prompt=prompt,
+                    response=response,
+                    input_metadata=input_metadata or {},
```
ulya-tkch, line 706: is input_metadata being validated anywhere?

cleanlab_studio/studio/trustworthy_language_model.py
ulya-tkch: I agree that TLMScoreResponse correctly belongs in types

cleanlab_studio/internal/api/api.py
huiwengoh: It is constructed/validated [here](https://github.com/cleanlab/cleanlab-studio/blob/94bcf094a12db7897eb87477f80e0cf03b875a4a/cleanlab_studio/studio/trustworthy_language_model.py#L519) (among other places where there are `kwargs` inputs)

cleanlab_studio/studio/trustworthy_language_model.py
huiwengoh: It was a TODO to validate the input log options, `log` options are already validated [here](https://github.com/cleanlab/cleanlab-studio/blob/94bcf094a12db7897eb87477f80e0cf03b875a4a/cleanlab_studio/internal/tlm/validation.py#L184-L193), so I'll delete this

cleanlab_studio/internal/tlm/validation.py
huiwengoh: This existing logic was actually confusing, updated the whole block

cleanlab_studio/internal/tlm/validation.py
huiwengoh: This existing logic was actually confusing, updated the whole block

cleanlab_studio/internal/api/api.py
ulya-tkch: I see it is treated as a kwarg. That part LGTM then!

ulya-tkch (approved):
Changes LGTM! Thanks for getting this in

cleanlab_studio/internal/tlm/validation.py
```diff
@@ -178,3 +181,68 @@ def validate_tlm_options(options: Any) -> None:
...
+                    raise ValidationError(
+                        f"Invalid type {type(val)}, perplexity should be a float when prompt is a str."
+                    )
+                if val is not None and not 0 <= val <= 1:
+                    raise ValidationError("Perplexity values must be between 0 and 1")
+
+            elif isinstance(prompt, Sequence):
+                if not isinstance(val, Sequence):
+                    raise ValidationError(
+                        f"Invalid type {type(val)}, perplexity should be a sequence when prompt is a sequence"
```
ulya-tkch, line 221: not sure if this is better but just a thought

```suggestion
                        f"Invalid type {type(val)}, perplexity should be a sequence of perplexity scores when prompt is a sequence of prompts."
```

cleanlab_studio/studio/trustworthy_language_model.py
jwmueller: but why isn't it instead defined as:

`TLMScoreResponse = Union[float, TLMScore]`

where `TLMScore` is defined as a TypedDict similar to `TLMResponse` and it has docstring like:

> Similar to TLMResponse dict, except without the response key.

?

jwmueller:
@huiwengoh Can you update PR description to also show the batched version of this?

huiwengoh:
Changes are happening in PR again after requested changes in the backend PR, so hold off from reviewing for now (@jas2600 )

cleanlab_studio/internal/tlm/validation.py
```diff
@@ -178,3 +181,70 @@ def validate_tlm_options(options: Any) -> None:
...
+                elif isinstance(response, Sequence):
+                    if not isinstance(val, Sequence):
+                        raise ValidationError(
+                            f"Invalid type {type(val)}, perplexity should be a sequence when response is a sequence"
+                        )
+                    if len(response) != len(val):
+                        raise ValidationError(
+                            "Length of the response and perplexity lists must match."
+                        )
+                    if not all(v is None or 0 <= v <= 1 for v in val):
```
ulya-tkch, line 229: if only part of the perplexity list is invalid such as `perplexity=[0, "bad_string"]`, then this will cause uncaught error. I'd recommend first checking each of the items in the list is 

```python
  if not (val is None or isinstance(val, float) or isinstance(val, int)):
      raise ValidationError(
          f"Invalid type {type(val)}, perplexity should be a float when response is a str."
      )
  if val is not None and not 0 <= val <= 1:
      raise ValidationError("Perplexity values must be between 0 and 1")
```

ulya-tkch (approved):
Code lgtm, one error not caught on client side but I left a comment. 
Thanks for getting this in :D

jas2600 (approved):

PR #239, opened 2024-06-14
add api methods for deploying model

axl1313:
Adds methods to `deploy_model()` and `wait_until_model_ready()` through Python API. These methods are required to enable us to add regression testing for model deployment.

### DIFF
--- a/cleanlab_studio/errors.py
+++ b/cleanlab_studio/errors.py
@@ -137,3 +137,7 @@ class InvalidSchemaTypeError(ValueError):
 
 class InvalidProjectConfiguration(HandledError):
     pass
+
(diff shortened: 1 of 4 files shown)


ryansingman (approved):
Changes LGTM. Not a huge fan of the code duplication between `poll_deployment_status` and [`poll_cleanset_status`](https://github.com/cleanlab/cleanlab-studio/blob/d001c59cb4b5f52a3936e0e6b7a05495ddd88909/cleanlab_studio/internal/clean_helpers.py#L13), but we can deduplicate once we add enrichment project polling (and have a better base on which to find the abstraction)

PR #240, opened 2024-06-17
Improvements to constrained outputs in DE

huiwengoh:
Note: `constrain_outputs` can be potentially further improved

### DIFF
--- a/cleanlab_studio/internal/enrichment_utils.py
+++ b/cleanlab_studio/internal/enrichment_utils.py
@@ -54,7 +54,7 @@ def get_regex_match_or_replacement(
         compiled_pattern = re.compile(regex)
         pattern_match = compiled_pattern.findall(response)
         if len(pattern_match) > 0:
-            return pattern_match[0]
+            return str(pattern_match[0])
         return None
 
     # if the regex is a tuple (or list of tuples), do replacement
--- a/cleanlab_studio/internal/enrichment_utils.py
+++ b/cleanlab_studio/internal/enrichment_utils.py
@@ -92,7 +92,7 @@ def get_optimized_prompt(prompt: str, constrain_outputs: Optional[List[str]] = N
 
     if constrain_outputs is not None:
         string_constrain_outputs = ", ".join(constrain_outputs)
-        pre_prompt = f"Your answer must only contain one of the following values: [{string_constrain_outputs}].\n"
+        pre_prompt = f"Your answer must exactly match one of the following values: [{string_constrain_outputs}].\n"
         optimal_prompt = f"{pre_prompt}{prompt}"
     else:
         optimal_prompt = prompt
--- a/cleanlab_studio/internal/enrichment_utils.py
+++ b/cleanlab_studio/internal/enrichment_utils.py
@@ -91,8 +91,8 @@ def get_optimized_prompt(prompt: str, constrain_outputs: Optional[List[str]] = N
     """
 
     if constrain_outputs is not None:
-        string_constrain_outputs = str(constrain_outputs).replace("'", "")
-        pre_prompt = f"Your answer must exactly match one of the following values: [{string_constrain_outputs}].\n"
+        string_constrain_outputs = ", ".join(constrain_outputs)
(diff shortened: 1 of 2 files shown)


cleanlab_studio/internal/enrichment_utils.py
```diff
@@ -107,11 +107,13 @@ def get_constrain_outputs_match(
 ) -> str:
     """Extracts the provided output values from the response using regex patterns. Return first extracted value if multiple exist.
     If no value out of the possible `constrain_outputs` is directly mentioned in the response, the return value with greatest string similarity to the response is returned (along with a warning).
+    If there are no close matches between the LLM response and any of the possible `constrain_outputs`, then the last entry of the `constrain_outputs` list is returned.
 
     Params
     ------
     response: Response from the LLM
-    constrain_outputs: List of expected output values
+    constrain_outputs: List of expected output values, the last value of this list should be considered the deafult/baseline value (eg. “other”),
```
mturk24, line 115:
```suggestion
    constrain_outputs: List of expected output values, the last value of this list should be considered the default/baseline value (eg. “other”),
```

cleanlab_studio/internal/enrichment_utils.py
```diff
@@ -91,8 +91,8 @@ def get_optimized_prompt(prompt: str, constrain_outputs: Optional[List[str]] = N
     """
 
     if constrain_outputs is not None:
-        string_constrain_outputs = str(constrain_outputs).replace("'", "")
-        pre_prompt = f"Your answer must exactly match one of the following values: [{string_constrain_outputs}].\n"
+        string_constrain_outputs = ", ".join(constrain_outputs)
+        pre_prompt = f"Your answer must only contain one of the following values: [{string_constrain_outputs}].\n"
```
mturk24, line 95: Do we still want to leave in the list wrapped around the string_constrain_outputs here?

huiwengoh: The current prompt template is:
```
Your answer must only contain one of the following values: [choice1, choice2].
```

Are you suggesting that we should remove the square braces all together? (this change so far has updated it from double square braces -> single)

mturk24: I think single list should be fine but also didn't test it without any brackets, I can always try without them locally though

cleanlab_studio/internal/enrichment_utils.py
```diff
@@ -52,9 +52,9 @@ def get_regex_match_or_replacement(
     # if the regex is a string, do matching
     if isinstance(regex, str):
         compiled_pattern = re.compile(regex)
-        pattern_match = compiled_pattern.match(response)
-        if pattern_match:
-            return pattern_match.group(1)
+        pattern_match = compiled_pattern.findall(response)
+        if len(pattern_match) > 0:
+            return str(pattern_match[0])
         return None
```
ulya-tkch, line 58: previously, there was discussion if this is best returned as None or the initial str. This would be a good PR to change that if necessary.

cleanlab_studio/internal/enrichment_utils.py
```diff
@@ -107,11 +107,13 @@ def get_constrain_outputs_match(
 ) -> str:
     """Extracts the provided output values from the response using regex patterns. Return first extracted value if multiple exist.
     If no value out of the possible `constrain_outputs` is directly mentioned in the response, the return value with greatest string similarity to the response is returned (along with a warning).
+    If there are no close matches between the LLM response and any of the possible `constrain_outputs`, then the last entry of the `constrain_outputs` list is returned.
```
ulya-tkch, line 110: This seems like confusing behavior to me that makes it hard to decide if the LLM is behaving badly or there are wrong outputs listed in contrain_outputs (think typos/capitalization?). I think the better behavior is to return np.nan and have the user handle this case (unless you've already considered this)

ulya-tkch: Nevermind, I see there is a preference to making the last value the "other".  I think that is ok then

ulya-tkch (approved):
Minor updates LGTM! I feel like keeping the list of possible constrained outputs between [] is best.

PR #241, opened 2024-06-17
Mypy fix - update np.float_ to np.float64

LukeMainwaring:
I'm updating this line to have more a modern and explicit type annotation for numpy floats. `np.float_` is removed in later NumPy versions


huiwengoh (approved):
Thanks for the quick fix!

PR #246, opened 2024-06-20
Add gpt-4o as TLM model option

huiwengoh:
Adds `gpt-4o` TLM model options, do not merge until backend is ready.


ryansingman (approved):
LGTM!

PR #251, opened 2024-06-26
Use response field 'trustworthiness_score' instead of 'trustworthy_sc…

eytanhanig:
The enrichment preview API endpoint previously returned the field `trustworthy_score`, however that field has since been renamed to `trustworthiness_score`. This updates the cleanlab-studio Python client to take that into account.

### DIFF
--- a/cleanlab_studio/studio/enrichment.py
+++ b/cleanlab_studio/studio/enrichment.py
@@ -19,7 +19,7 @@ from cleanlab_studio.studio.trustworthy_language_model import TLMOptions
 Replacement = Tuple[str, str]
 ROW_ID_COLUMN_NAME = "row_id"
 FINAL_RESULT_COLUMN_NAME = "final_result"
-TRUSTWORTHY_SCORE_COLUMN_NAME = "trustworthiness_score"
+TRUSTWORTHINESS_SCORE_COLUMN_NAME = "trustworthiness_score"
 LOG_COLUMN_NAME = "log"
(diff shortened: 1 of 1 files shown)


cleanlab_studio/studio/enrichment.py
```diff
@@ -19,7 +19,7 @@
 Replacement = Tuple[str, str]
 ROW_ID_COLUMN_NAME = "row_id"
 FINAL_RESULT_COLUMN_NAME = "final_result"
-TRUSTWORTHY_SCORE_COLUMN_NAME = "trustworthy_score"
+TRUSTWORTHY_SCORE_COLUMN_NAME = "trustworthiness_score"
```
ulya-tkch, line 22: Should we update the variable to also be similar here? Might cause confusion if its referenced as both the trustworthy score and trustworthiness score

```suggestion
TRUSTWORTHINESS_SCORE_COLUMN_NAME = "trustworthiness_score"
```

eytanhanig: Good call. Change made.

ulya-tkch (approved):
LGTM! Make sure the parameter is not imported elsewhere in the codebase and if it is, naming is changed. 

Small refactor for later: these constants should live in `constants.py` and be imported from there [like so](https://github.com/cleanlab/cleanlab-studio/blob/50e4867e1246b8c0e17afd32a53c62de7177c819/cleanlab_studio/studio/trustworthy_language_model.py#L30)

PR #264, opened 2024-07-18
Add TLMHybrid class

huiwengoh:
Adding a TLMHybrid class that allows prompting one LLM for getting responses, and another LLM to get the trustworthiness scores.

Sample usage:
```python
tlm_hybrid = studio.TLMHybrid("gpt-4o")
tlm_hybrid.prompt("What do penguins eat?")

# batch prompting
tlm_hybrid.prompt(["what is 1+1?", "what is 2+2?"])
```


cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -616,6 +619,276 @@ async def _get_trustworthiness_score_async(
...
+        Use `Studio.TLMHybrid()` instead of this method to initialize a TLMHybrid object.
+        lazydocs: ignore
+        """
+        self._api_key = api_key
+        self._response_model = response_model
+
+        if quality_preset in {"high", "best"}:
+            raise ValidationError(
+                f"Invalid quality preset: {quality_preset}. TLMHybrid only supports 'low' and 'medium' presets."
+            )
```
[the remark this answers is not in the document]
huiwengoh: The quality preset here specifies the preset used in the `get_trustworthiness_score()` call, for which `medium`/`high`/`best` presets do the same thing (the difference between `high` and `best` is the number of candidate responses which is only relevant for `prompt()`).
So instead of offering options which don't actually change the quality of the score at all (it will be confusing(, we are just disallowing it.

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -616,6 +619,276 @@ async def _get_trustworthiness_score_async(
...
+        self._score_quality_preset = quality_preset
+
+        if options is not None:
+            validate_tlm_hybrid_score_options(options)
+        self._score_options = options
+        self._prompt_options = cast(
+            TLMOptions, get_tlm_hybrid_prompt_options(self._score_options, self._response_model)
+        )
+
+        self._timeout = timeout  # TODO: are we handling this any differently?
```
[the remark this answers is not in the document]
huiwengoh: I just had one consideration here: would it make sense that each call gets terminated halfway then?
eg. if I specify `timeout=10` would it be strange if I get an error in the `prompt` step after 5 seconds?

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -616,6 +619,276 @@ async def _get_trustworthiness_score_async(
...
+            Models currently supported include "gpt-3.5-turbo-16k", "gpt-4", "gpt-4o", "claude-3-haiku".
+
+        quality_preset (TLMQualityPreset, default = "medium"): preset configuration to control the quality of TLM trustworthiness scores vs. runtimes/costs.
+            Note that this preset only applies to the model computing the trustworthiness score.
+            Supported options include: "medium", "low".
+
+        options (TLMOptions, optional): a typed dict of advanced configuration options.
+            Note that most of these options only apply to the model computing the trustworthiness score, except for "max_tokens", which applies to both models.
+            Available options (keys in this dict) include "model", "max_tokens", "num_consistency_samples", "use_self_reflection".
+            For more details about the options, see the documentation for [TLMOptions](#class-tlmoptions).
```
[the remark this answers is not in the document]
jwmueller: I think TLMOptions (which is already a lot) will become too complex if we try to also add in TLMHybrid comments there.

Instead, TLMHybrid documentation should say what parts of TLMOptions may be differently applied in the hybrid case.

You should always think of a TLMHybrid user as somebody who's already used TLM, is happy with the results, and just wants to get lower latency / costs. So they don't need to know too many specifics, as long as they know what all the options are to experiment with and which direction to tweak these options to reduce latency/costs

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -616,6 +619,276 @@ async def _get_trustworthiness_score_async(
             raise e
 
 
+class TLMHybrid:
```
jwmueller, line 622: How about we move all the Hybrid specific stuff into new file like `trustworthy_language_model.hybrid`, WDYT?

I think it'd lead to less confusion/clutter since TLM and TLMHybrid have most of the same methods

cleanlab_studio/internal/tlm/validation.py
```diff
@@ -255,3 +255,33 @@ def process_response_and_kwargs(
...
+
+def validate_tlm_hybrid_score_options(score_options: Any) -> None:
+    validate_tlm_options(score_options)
+
+    INVALID_SCORE_OPTIONS = {"num_candidate_responses"}
+
+    invalid_score_keys = set(score_options.keys()).intersection(INVALID_SCORE_OPTIONS)
+    if invalid_score_keys:
+        raise ValidationError(
+            f"Invalid keys in options dictionary for TLMHybrid: {invalid_score_keys}.\n"
```
jwmueller, line 268:
```suggestion
            f"Please remove these invalid keys from the `options` dictionary provided for TLMHybrid: {invalid_score_keys}.\n"
```

cleanlab_studio/internal/tlm/validation.py
```diff
@@ -255,3 +255,33 @@ def process_response_and_kwargs(
...
+        )
+
+
+def get_tlm_hybrid_prompt_options(score_options: Any, response_model: str) -> Dict[str, Any]:
+    VALID_PROMPT_OPTIONS = {"max_tokens"}
+
+    if response_model not in _VALID_TLM_MODELS:
+        raise ValidationError(
+            f"{response_model} is not a supported model, valid models include: {_VALID_TLM_MODELS}"
+        )
```
jwmueller, line 278: why do we need to do this check? Isn't the response_model going to be validated elsewhere when we internally instantiate a base-preset TLM out of it?

cleanlab_studio/internal/tlm/validation.py
```diff
@@ -255,3 +255,33 @@ def process_response_and_kwargs(
...
+
+    prompt_options = {"model": response_model, "log": ["perplexity"]}
```
jwmueller, line 280: just curious: why is this called `prompt_options` instead of `response_options`

cleanlab_studio/studio/studio.py
```diff
@@ -436,6 +436,29 @@ def TLM(
...
+        self,
+        response_model: str,
+        quality_preset: TLMQualityPreset = "medium",
+        *,
+        options: Optional[trustworthy_language_model.TLMOptions] = None,
+        timeout: Optional[float] = None,
+        verbose: Optional[bool] = None,
+    ) -> trustworthy_language_model.TLMHybrid:
+        """
+        Instantiate a TLMHybrid instance.
```
jwmueller, line 449:
```suggestion
        Instantiate a hybrid Trustworthy Language Model that uses one model for response and another for trustworthiness scoring.
```

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -616,6 +619,276 @@ async def _get_trustworthiness_score_async(
...
+    ** The TLMHybrid object is not meant to be constructed directly.** Instead, use the [`Studio.TLMHyrid()`](../studio/#method-tlmhyrid)
+    method to configure and instantiate a TLMHybrid object.
+    After you've instantiated the TLMHybrid object using [`Studio.TLMHybrid()`](../studio/#method-tlmhyrid), you can use the instance methods documented on this page.
+    Possible arguments for `Studio.TLMHybrid()` are documented below.
+
+    Most of the input arguments for this class are similar to those for TLM, major differences will be described below.
+
+    Args:
+        response_model (str): LLM used to produce the response to the given prompt.
+            Do not specify the model to use for obtaining the trustworthiness score here, instead specify it in the options argument.
```
jwmueller, line 638:
```suggestion
            Do not specify the model to use for scoring  trustworthiness here, instead specify that model in the `options` argument.
```

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -616,6 +619,276 @@ async def _get_trustworthiness_score_async(
...
+            Models currently supported include "gpt-3.5-turbo-16k", "gpt-4", "gpt-4o", "claude-3-haiku".
```
jwmueller, line 639:
```suggestion
            The list of supported model strings can be found in the TLMOptions documentation.
```

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -616,6 +619,276 @@ async def _get_trustworthiness_score_async(
...
+
+        quality_preset (TLMQualityPreset, default = "medium"): preset configuration to control the quality of TLM trustworthiness scores vs. runtimes/costs.
+            Note that this preset only applies to the model computing the trustworthiness score.
+            Supported options include: "medium", "low".
```
jwmueller, line 643:
```suggestion
            This preset only applies to the model computing the trustworthiness score.
            Supported options are only: "medium" or "low",
            because TLMHybrid is not intended to improve response accuracy (use the regular TLM for that).
```

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -616,6 +619,276 @@ async def _get_trustworthiness_score_async(
             raise e
 
 
+class TLMHybrid:
+    """
+    A hybrid version of the Trustworthy Language Model (TLM) that enables the use of different LLMs for the response and for obtaining the trustworthiness score.
```
jwmueller, line 624:
```suggestion
    A hybrid version of the Trustworthy Language Model (TLM) that enables the use of different LLMs for the response and for scoring its trustworthiness.
```

cleanlab_studio/studio/studio.py
```diff
@@ -436,6 +436,29 @@ def TLM(
             self._api_key, quality_preset, options=options, timeout=timeout, verbose=verbose
         )
 
+    def TLMHybrid(
+        self,
+        response_model: str,
```
jwmueller, line 441:
```suggestion
        response_model: str = "gpt-4o",
```

jwmueller: Let's have a default, so this whole thing works with: 
`tlm = studio.TLMHybrid()`

Our tutorial should still explicitly demonstrate setting specific response/scoring model

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -616,6 +619,276 @@ async def _get_trustworthiness_score_async(
...
+        response_model (str): LLM used to produce the response to the given prompt.
+            Do not specify the model to use for obtaining the trustworthiness score here, instead specify it in the options argument.
+            Models currently supported include "gpt-3.5-turbo-16k", "gpt-4", "gpt-4o", "claude-3-haiku".
+
+        quality_preset (TLMQualityPreset, default = "medium"): preset configuration to control the quality of TLM trustworthiness scores vs. runtimes/costs.
+            Note that this preset only applies to the model computing the trustworthiness score.
+            Supported options include: "medium", "low".
+
+        options (TLMOptions, optional): a typed dict of advanced configuration options.
+            Note that most of these options only apply to the model computing the trustworthiness score, except for "max_tokens", which applies to both models.
```
jwmueller, line 646:
```suggestion
            Most of these options only apply to the model scoring  trustworthiness, except for "max_tokens", which applies to the response model as well.
```

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -616,6 +619,276 @@ async def _get_trustworthiness_score_async(
...
+            Available options (keys in this dict) include "model", "max_tokens", "num_consistency_samples", "use_self_reflection".
```
jwmueller, line 647:
```suggestion
```

jwmueller: we want to avoid listing options out as much as possible across TLMHybrid documentation, so we don't need to keep updating it

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -616,6 +619,276 @@ async def _get_trustworthiness_score_async(
...
+            options=self._score_options,
+            timeout=self._timeout,
+            verbose=self._verbose,
+        )
+
+    def prompt(
+        self,
+        prompt: Union[str, Sequence[str]],
+    ) -> Union[TLMResponse, List[TLMResponse]]:
+        """
```
jwmueller, line 708: Instead of all this could we just point the reader to the TLM prompt() documentation?

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -616,6 +619,276 @@ async def _get_trustworthiness_score_async(
...
+            TLMResponse | List[TLMResponse]: [TLMResponse](#class-tlmresponse) object containing the response and trustworthiness score.
+                If multiple prompts were provided in a list, then a list of such objects is returned, one for each prompt.
+                This method will raise an exception if any errors occur or if you hit a timeout (given a timeout is specified).
+                Use it if you want strict error handling and immediate notification of any exceptions/timeouts.
+
+                If running this method on a big batch of prompts: you might lose partially completed results if TLM fails on any one of them.
+                To avoid losing partial results for the prompts that TLM did not fail on,
+                you can either call this method on smaller batches of prompts at a time
+                (and save intermediate results between batches), or use the [`try_prompt()`](#method-try_prompt) method instead.
+        """
```
jwmueller, line 727: Instead of all this could we just point the reader to the TLM prompt() documentation?

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -616,6 +619,276 @@ async def _get_trustworthiness_score_async(
...
+        prompt_response = self.tlm_response.prompt(prompt)
+
+        if isinstance(prompt_response, list):
+            assert isinstance(prompt, Sequence)
```
jwmueller, line 731: why is there a test here?

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -616,6 +619,276 @@ async def _get_trustworthiness_score_async(
...
+            response = [r["response"] for r in prompt_response]
+            perplexity = [r["log"]["perplexity"] for r in prompt_response]
+            return self._batch_score(prompt, response, perplexity)
+
+        assert isinstance(prompt, str)
```
jwmueller, line 736: why is there a test here?

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -616,6 +619,276 @@ async def _get_trustworthiness_score_async(
...
+                To avoid losing partial results for the prompts that TLM did not fail on,
+                you can either call this method on smaller batches of prompts at a time
+                (and save intermediate results between batches), or use the [`try_prompt()`](#method-try_prompt) method instead.
+        """
+        prompt_response = self.tlm_response.prompt(prompt)
+
+        if isinstance(prompt_response, list):
+            assert isinstance(prompt, Sequence)
+            response = [r["response"] for r in prompt_response]
+            perplexity = [r["log"]["perplexity"] for r in prompt_response]
```
jwmueller, line 733: what if the response model doesn't support perplexity, like claude?

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -616,6 +619,276 @@ async def _get_trustworthiness_score_async(
...
+        Returns:
+            List[Optional[TLMResponse]]: list of [TLMResponse](#class-tlmresponse) objects containing the response and trustworthiness score.
+                The returned list will always have the same length as the input list.
+                In case of TLM failure on any prompt (due to timeouts or other errors),
+                the return list will contain None in place of the TLM response for that failed prompt.
+                Use this to obtain TLM results for as many prompts as possible,
+                but you might miss out on certain error messages.
+                If you prefer to be notified immediately about any errors or timeouts when running many prompts,
+                use the [`prompt()`](#method-prompt) method instead.
+        """
```
jwmueller, line 768: Instead of all this could we just point the reader to the TLM prompt() documentation?

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -616,6 +619,276 @@ async def _get_trustworthiness_score_async(
...
+        response_succeeded = [r["response"] for r in prompt_response_succeeded]
+        perplexity_succeeded = [r["log"]["perplexity"] for r in prompt_response_succeeded]
+        score_response_succeeded = self._try_batch_score(
+            prompt_succeeded, response_succeeded, perplexity_succeeded
+        )
+
+        tlm_response = np.full(len(prompt), None)
+        tlm_response[prompt_succeeded_mask] = score_response_succeeded
+
+        return cast(List[Optional[TLMResponse]], tlm_response.tolist())
```
jwmueller, line 788: could you just confirm that the returned values look exactly the same as from TLM.try_prompt() in batch-settings?

jwmueller (changes requested):
looking good, but had some suggestions

Could be helpful for testing if you can also simultaneously PR your tutorial for using TLMHybrid for our docs

cleanlab_studio/studio/trustworthy_language_model.py
huiwengoh: This was required for mypy to pass. Because the input prompt arg can either be a sequence or str, so we needed to make sure that it was of the right type for each case.

Updated these to be explicit if/else statements.

cleanlab_studio/studio/trustworthy_language_model.py
huiwengoh: It will still work! 
If the model does not support perplexity, `r["log"]["perplexity"]` will be `None`, and the `perplexity` kwarg in `get_trustworthiness_score()` will also accept `None` as an input (it will just not be used and the scores will be computed as usual).

cleanlab_studio/studio/trustworthy_language_model.py
jwmueller: could be: `utils.tlm_hybrid`

cleanlab_studio/internal/tlm/validation.py
huiwengoh: In my head it was "the options used in the `prompt` method", hence `prompt_options`, but I agree that `response_options` matches the naming convention better (given `response_model`), can update

cleanlab_studio/studio/trustworthy_language_model.py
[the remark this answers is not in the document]
huiwengoh: > We should be able to make it precise by tracking the elapsed time for the first call and then set the remaining time as the timeout for the second call.

It unfortunately won't be as simple as this, as for batch prompts the timeout specified is applied per-prompt, not on the entire batch (ie. the entire batch call would likely take longer than the specified `timeout`).

I will keep as if for now, and add a TODO for future improvement.

cleanlab_studio/studio/trustworthy_language_model.py
huiwengoh: Moved it into `utils.tlm_hybrid`! Docs look like this now:
image

The only thing I can't fix is the **Tlm Hybrid** titile. It seems like the capitalization is automatic when the file is converted to md (I can't find a way for it to capitalize TLM in the configs), the only way I can fix it is to write a script to edit the MD file generated - is that something that we should do (feels quite hacky)?

cleanlab_studio/studio/trustworthy_language_model.py
huiwengoh: Yup! It will be a `TLMResponse` dict with both response and score if it succeeds, and `None` if either step fails.

cleanlab_studio/studio/trustworthy_language_model.py
jwmueller: thanks this seems good. No worries about the capitalization, this is ok for now

jwmueller (approved):
LGTM on the surface-level.

I didn't play with it in detail, leave that to @jas2600 .
Please make sure when playing with it to determine user-experience is good, that you try using openai model for response and claude for scoring and vice-versa

jas2600:
It looks good overall! Just two minor things:

- I think we should include the `base` preset in `Invalid quality preset: {quality_preset}. TLMHybrid only supports 'low' and 'medium' presets.` as well. Currently it's `Cannot get trustworthiness score with `base` quality_preset -- choose a higher preset.`
- When given invalid preset like `test` it gives `Invalid quality preset test -- must be one of ['best', 'high', 'medium', 'low', 'base']`. It should be `must be one of ['medium', 'low']`

Screenshot 2024-07-23 at 7 44 14 PM
Screenshot 2024-07-23 at 7 44 05 PM

Screenshot 2024-07-23 at 7 44 53 PM
Screenshot 2024-07-23 at 7 44 37 PM
Screenshot 2024-07-23 at 7 45 01 PM
Screenshot 2024-07-23 at 7 44 26 PM


Screenshot 2024-07-23 at 8 12 02 PM
Screenshot 2024-07-23 at 7 45 18 PM
Screenshot 2024-07-23 at 7 45 09 PM

cleanlab_studio/utils/tlm_hybrid.py
```diff
@@ -0,0 +1,255 @@
...
+        Use `Studio.TLMHybrid()` instead of this method to initialize a TLMHybrid object.
+        lazydocs: ignore
+        """
+        self._api_key = api_key
+        self._response_model = response_model
+
+        if quality_preset in {"high", "best"}:
+            raise ValidationError(
+                f"Invalid quality preset: {quality_preset}. TLMHybrid only supports 'low' and 'medium' presets."
+            )
```
jas2600, line 73: Should include `base`

huiwengoh: Updated!

jas2600 (approved):

PR #265, opened 2024-07-18
docs fix: remove api_key arg from docstring of TLM

jwmueller:
This is not a valid arg, think it got accidentally copied in the docstring when moving from studio -> trustworthy_language_model module

Screen Shot 2024-07-17 at 9 05 27 PM

### DIFF
--- a/cleanlab_studio/studio/trustworthy_language_model.py
+++ b/cleanlab_studio/studio/trustworthy_language_model.py
@@ -47,8 +47,6 @@ class TLM:
     After you've instantiated the TLM object using [`Studio.TLM()`](../studio/#method-tlm), you can use the instance methods documented on this page. Possible arguments for `Studio.TLM()` are documented below.
 
     Args:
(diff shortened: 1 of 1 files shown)


jwmueller:
@jas2600 after you merge this, could you release new version of the cleanlab-studio client, so we can get this bugfix live on the docs?

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -47,8 +47,6 @@ class TLM:
     After you've instantiated the TLM object using [`Studio.TLM()`](../studio/#method-tlm), you can use the instance methods documented on this page. Possible arguments for `Studio.TLM()` are documented below.
 
     Args:
-        api_key: You can find your API key on your [account page](https://app.cleanlab.ai/account) in Cleanlab Studio. Instead of specifying the API key here, you can also log in with `cleanlab login` on the command-line.
```
jas2600, line 50: The TLM class itself does require `api_key` though. But agree to put it here might be confusing since the studio.TLM() function doesn't require and we hide the __init__ of this class.

jas2600 (approved):

jas2600:
Screenshot 2024-07-18 at 10 36 13 AM

@jwmueller the doc fix is live now

PR #278, opened 2024-07-30
[bugfix] "ModuleNotFoundError: No module named 'snowflake.snowpark'" that happens when running Studio API tutorials

jwmueller:
Once this bugfix is merged in, please ask @aditya1503 why #benchmark-studio regression tests did not catch this.

This is high-priority bug that affects API users.

Screen Shot 2024-07-29 at 7 30 36 PM

### DIFF
--- a/cleanlab_studio/internal/api/api.py
+++ b/cleanlab_studio/internal/api/api.py
@@ -26,6 +26,7 @@ from io import StringIO
 
 try:
     import snowflake
+    import snowflake.snowpark as snowpark
 
     snowflake_exists = True
 except ImportError:


jwmueller:
please also verify there are no other lazy/optional imports like this where the `try` statement doesn't match what is later actually imported (eg. for databricks)

cleanlab_studio/internal/api/api.py
```diff
@@ -26,6 +26,7 @@
 
 try:
     import snowflake
+    import snowflake.snowpark as snowpark
```
jwmueller, line 29: note I didn't verify this update will actually suffice to make the snowflake tutorial run properly.  But it should at least prevent all the other API tutorials from being broken like they currently are.

The snowflake tutorial is also broken in other ways:
https://github.com/cleanlab/cleanlab-studio-docs/issues/411

If you end up looking into it, please push other fixes for that tutorial too (and also push them for the databricks tutorial which has similar breakage).

jwmueller: my guess is the v0.11 release of [snowflake package](https://pypi.org/project/snowflake/0.11.0/) on July 25 broke this import. 

So either: this needs to be updated to the new import with a lower-bound version of snowflake for our tutorial, or we upper-bound the version of snowflake to v0.10
our tutorial needs to version-cap snowflake, or best is if you find an easy fix that works for all snowflake versions

ulya-tkch:
What are the steps to reproduce this bug @jwmueller?
I tried to follow the installs/imports for the snowflake tutorial and did not run into any issues importing cleanlab-studio in the later cells. 

```python3
# create new virtual env
# python3 -m venv ~/virtual/bugfix
# source ~/virtual/bugfix/bin/activate

%pip install openai
%pip install snowflake-snowpark-python
%pip install --upgrade cleanlab-studio

import openai
from openai import OpenAI
import os
from snowflake.snowpark import Session
import snowflake.snowpark.functions as F
import cleanlab_studio
```

In fact, it seems both %pip install snowflake and %pip install snowflake-snowpark-python work for me in above ^

jwmueller:
> What are the steps to reproduce this bug @jwmueller?
> 
> I tried to follow the installs/imports for the snowflake tutorial and did not run into any issues importing cleanlab-studio in the later cells. 
> 
> 
> 
> ```python3
> 
> # create new virtual env
> 
> # python3 -m venv ~/virtual/bugfix
> 
> # source ~/virtual/bugfix/bin/activate
> 
> 
> 
> %pip install openai
> 
> %pip install snowflake-snowpark-python
> 
> %pip install --upgrade cleanlab-studio
> 
> 
> 
> import openai
> 
> from openai import OpenAI
> 
> import os
> 
> from snowflake.snowpark import Session
> 
> import snowflake.snowpark.functions as F
> 
> import cleanlab_studio
> 
> ```
> 
> 
> 
> In fact, it seems both %pip install snowflake and %pip install snowflake-snowpark-python work for me in above ^
> 
> 
> 
> Oh it is only replicable in colab, right. Will work on that then. AFAIK, jupyter notebook is not affected by this then

Yes the general API failing upon import is only in colab, because there something is already be preinstalled with regards to snowflake. 

That said I'm pretty such our snowflake integration code no longer works with snowflake v0.11 although I haven't verified

ulya-tkch (approved):
Studio tutorials are runnable in Colab with this fix. Should get this merged and I will separate the other tasks into issues separately.

ulya-tkch:
The only other lazy import is pyspark.sql which is handled appropriately.

PR #280, opened 2024-07-30
[update version] to account for tutorial bugflx

ulya-tkch:
Update version to account for Studio tutorials bugfix pr


huiwengoh (changes requested):
I think you are on some outdated version of this file. The latest version is `2.1.7` ([here](https://github.com/cleanlab/cleanlab-studio/blob/main/cleanlab_studio/version.py#L4)).

huiwengoh (approved):

PR #281, opened 2024-07-30
cleanlab studio beta api

axl1313:


cleanlab_studio/internal/api/api_helper.py
```diff
@@ -10,3 +26,37 @@ def check_uuid_well_formed(uuid_string: str, id_name: str) -> None:
...
+    retval = dict()
+    if api_key:
+        retval["Authorization"] = f"bearer {api_key}"
+    if content_type:
+        retval["Content-Type"] = content_type
+    retval["Client-Type"] = "python-api"
+    return retval
+
+
+def handle_api_error(res: requests.Response) -> None:
```
ulya-tkch, line 43: Why is the code duplicated [here](https://github.com/cleanlab/cleanlab-studio/blob/49e9e3ffbe3c18c657c7cc343ed250fe9a426159/cleanlab_studio/cli/api_service.py#L39)?

PR #284, opened 2024-08-01
Add method to print TLM model name

huiwengoh:
Adds new method `get_model_name()` to easily know which model the TLM is using.

I've also explicitly specified which default model is being used in the client API call to the backend. Currently, if no model is specified from the client, we use the default in the backend but this `get_model_name()` method can get outdated if someone uses an older version of the client library against an updated backend.

Sample usage:

```python
tlm = studio.TLM()
tlm.get_model_name()

>> 'gpt-4o-mini'
```

```python
tlm = studio.TLM(options={"model": "claude-3-haiku"})
tlm.get_model_name()

>> 'claude-3-haiku'
```

```python
tlm_hybrid = studio.TLMHybrid()
tlm_hybrid.get_model_names()

>> {'response_model': 'gpt-4o', 'score_model': 'gpt-4o-mini'}
```

### DIFF
--- a/cleanlab_studio/studio/trustworthy_language_model.py
+++ b/cleanlab_studio/studio/trustworthy_language_model.py
@@ -98,10 +98,16 @@ class TLM:
             )
 
         self._return_log = False
-        if options is not None:
-            validate_tlm_options(options)
-            if "log" in options.keys():
-                self._return_log = True
+
+        options_dict = options or {}
+        validate_tlm_options(options_dict)
+        if "log" in options_dict.keys():
+            self._return_log = True
+
+        # explicitly specify the default model
+        self._options = {**{"model": _TLM_DEFAULT_MODEL}, **options_dict}
+
+        self._quality_preset = quality_preset
 
         if timeout is not None and not (isinstance(timeout, int) or isinstance(timeout, float)):
             raise ValidationError("timeout must be a integer or float value")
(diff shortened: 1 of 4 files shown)


cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -616,6 +621,10 @@ async def _get_trustworthiness_score_async(
                 return None
             raise e
 
+    def get_model_name(self) -> str:
+        """Returns the underlying LLM used to obtain responses and scoring trustworthiness."""
```
jwmueller, line 625:
```suggestion
        """Returns the underlying LLM used to generate responses and score their  trustworthiness."""
```

cleanlab_studio/utils/tlm_hybrid.py
```diff
@@ -253,3 +253,10 @@ def _try_batch_score(
             ]
         else:
             raise ValueError(f"score_response has invalid type")
+
+    def get_model_names(self) -> Dict[str, str]:
+        """Returns the underlying LLMs used to obtain responses and scoring trustworthiness."""
```
jwmueller, line 258:
```suggestion
        """Returns the underlying LLMs used to generate responses and score their trustworthiness."""
```

jwmueller (approved):
LGTM, just had minor suggestion

PR #286, opened 2024-08-07
[DE] implement `wait_until_ready` function for `EnrichmentProject`

LukeMainwaring:
The `wait_until_ready` function is designed to wait until the latest enrichment job is completed. It uses a progress bar to provide visual feedback on the job's progress.
- `num_rows` is the total # rows for the dataset and comes from the enrichment project dict. `processed_rows` is the value of currently processed rows and comes from the enrichment job status endpoint. these are used in the progress. the loop continues until `self.ready` is `True`
- `_get_num_processed_rows` helper function, because the status endpoint only returns the `processed_rows` property if the status is running. note: we may want to include this logic in the backend instead of python client
- refactored `self._latest_populate_job` to use `_get_latest_job` helper function. the previous implementation was problematic because it fetched the latest job of all types, including preview jobs, which made the naming misleading. maintaining an instance variable required constant updates across various helper functions, making the code harder to manage than fetching the latest job directly from the backend 

https://github.com/user-attachments/assets/40cd90f8-0fa5-48f6-a097-b86524fc552c

### DIFF
--- a/cleanlab_studio/studio/enrichment.py
+++ b/cleanlab_studio/studio/enrichment.py
@@ -192,7 +192,7 @@ class EnrichmentProject:
         )
         return response
 
-    def _get_latest_job(self) -> dict:
+    def _get_latest_job(self) -> EnrichmentJob:
         """Retrieve the latest job and its details."""
         return self.list_all_jobs()[0]
 
--- a/cleanlab_studio/studio/enrichment.py
+++ b/cleanlab_studio/studio/enrichment.py
@@ -33,7 +33,6 @@ REGEX_PARAMETER_ERROR_MESSAGE = (
     "The 'regex' parameter must be a string, a tuple(str, str), or a list of tuple(str, str)."
 )
 CLEANLAB_ROW_ID_COLUMN_NAME = "cleanlab_row_ID"
-CHECK_READY_INTERVAL = 60 * 2  # 2 minutes
 
 
 class EnrichmentJobStatusEnum(Enum):
--- a/cleanlab_studio/studio/enrichment.py
+++ b/cleanlab_studio/studio/enrichment.py
@@ -33,6 +33,7 @@ REGEX_PARAMETER_ERROR_MESSAGE = (
     "The 'regex' parameter must be a string, a tuple(str, str), or a list of tuple(str, str)."
 )
 CLEANLAB_ROW_ID_COLUMN_NAME = "cleanlab_row_ID"
+CHECK_READY_INTERVAL = 100
 
 
 class EnrichmentJobStatusEnum(Enum):
@@ -257,7 +258,7 @@ class EnrichmentProject:
                 pbar.set_postfix_str(latest_job_status["status"])
                 pbar.update(num_processed_rows - pbar.n)
 
-                for _ in range(50):
+                for _ in range(CHECK_READY_INTERVAL):
                     time.sleep(0.1)
                     pbar.set_description_str(f"Enrichment Progress: {next(spinner)}")
 
--- a/cleanlab_studio/studio/enrichment.py
+++ b/cleanlab_studio/studio/enrichment.py
@@ -214,6 +214,8 @@ class EnrichmentProject:
         """Check if the latest populate job is ready."""
         latest_job = self._get_latest_job()
         if latest_job["job_type"] != "ENRICHMENT":
+            # TODO: consider fetching latest populate job directly instead of throwing an error
+            # This would prevent the user from getting stuck in an error state if preview job is the latest job
             raise ValueError(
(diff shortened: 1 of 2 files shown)


cleanlab_studio/studio/enrichment.py
```diff
@@ -184,38 +192,87 @@ def run(
...
+                num_processed_rows = self._get_num_processed_rows()
+                if pbar.total is None and num_rows is not None:
+                    pbar.total = num_rows
+                    pbar.refresh()
+
+                pbar.set_postfix_str(latest_job_status["status"])
+                pbar.update(num_processed_rows - pbar.n)
+
+                for _ in range(50):
+                    time.sleep(0.1)
```
novking, line 261: why not just `sleep(50)`? and i think 2mins should be a sweet spot since this is average time for tlm to process data

cleanlab_studio/studio/enrichment.py
```diff
@@ -184,38 +192,87 @@ def run(
...
-        self.latest_populate_job = latest_job
-        if latest_job["status"] == "FAILED":
+        status = latest_job["status"]
+        if status == EnrichmentJobStatusEnum.FAILED.value:
             raise ValueError("The latest populate job failed.")
-        elif latest_job["status"] == "RUNNING":
-            return False
-        elif latest_job["status"] == "CREATED":
+        elif status in {
+            EnrichmentJobStatusEnum.RUNNING.value,
```
novking, line 232: thanks for making it as Enum. love it <3

cleanlab_studio/studio/enrichment.py
```diff
@@ -184,38 +192,87 @@ def run(
...
+        """Retrieve the latest job and its details."""
+        return self.list_all_jobs()[0]
+
+    def _get_latest_job_status(self) -> dict[str, Any]:
+        """Retrieve the latest job status with details, e.g. processed_rows and average_trustworthiness_score."""
+        latest_job = self._get_latest_job()
+        return api.get_enrichment_job_status(self._api_key, job_id=latest_job["id"])
+
+    def _get_num_processed_rows(self) -> int:
+        num_rows = int(self._get_enrichment_project_dict()["num_rows"])
```
novking, line 205: if "num_rows" saved in project. I think it's much better to do this on the backend since it would also save a call for UI as well.
For tmw demo, i think this is fine. but can you add a TODO for this. should be a simple iteration 🥇

novking (approved):
Good job.
You are right about the latest_job, i think even the PRD didn't have every cases mapped out.
I think the logic should be:
1. backend api: if there's a `populate` job, it will block all in-coming calls. including both `preview` and `populate`. i think right now, it's only block `populate`
2. backend api: we could potentially have the `list_job` api to take in the `type` param?
3. python client: if a user did a `preview` AFTER `populate`. then `ready` and other function should raise error and let user know that latest status is that you did a `preview` instead of a `populate`. if you want to retrieve the latest `populate` job. do something like `get_result(latest_enrichment_job=True)` or things like that.

cleanlab_studio/studio/enrichment.py
LukeMainwaring: the purpose of this code is create a progress indicator with a spinner animation, rather than just pausing for X seconds. total sleep time here will be 50 * 0.1 = 5 seconds. we'll want to poll more frequently to dynamically get progress percentages, but we can configure the proper polling frequency. i've re-introduced `CHECK_READY_INTERVAL` with a value of 100, so total sleep time in each while loop iteration is now 10 seconds

cleanlab_studio/studio/enrichment.py
LukeMainwaring: yes, i was planning on moving this logic to the backend instead of client-side, so i'll go ahead and make that change now while it's at the top of my mind

LukeMainwaring:
> Good job. You are right about the latest_job, i think even the PRD didn't have every cases mapped out. I think the logic should be:
> 
> 1. backend api: if there's a `populate` job, it will block all in-coming calls. including both `preview` and `populate`. i think right now, it's only block `populate`
> 2. backend api: we could potentially have the `list_job` api to take in the `type` param?
> 3. python client: if a user did a `preview` AFTER `populate`. then `ready` and other function should raise error and let user know that latest status is that you did a `preview` instead of a `populate`. if you want to retrieve the latest `populate` job. do something like `get_result(latest_enrichment_job=True)` or things like that.

The current code does throw an error if latest job is a `preview` instead of a full `run`, but now that i'm thinking more about it, it's not the best UX for the `ready` function to stick them in an error state if `preview` occurs after `run`. I'll think about a cleaner UX, but maybe it's similar to what you said, where we just "fetch the latest full enrichment job"
Screenshot 2024-08-07 at 5 02 02 PM

PR #287, opened 2024-08-14
[TLM] Handle empty log list

huiwengoh:
Does not try to return log if an empty list is passed into the `log` option. The following should work properly now:

```python
tlm = studio.TLM(options={"log": []})
res = tlm.prompt("What are the colours of a rainbow?")
```

### DIFF
--- a/cleanlab_studio/studio/trustworthy_language_model.py
+++ b/cleanlab_studio/studio/trustworthy_language_model.py
@@ -101,7 +101,7 @@ class TLM:
 
         options_dict = options or {}
         validate_tlm_options(options_dict)
-        if "log" in options_dict.keys():
+        if "log" in options_dict.keys() and len(options_dict["log"]) > 0:
             self._return_log = True
 
         # explicitly specify the default model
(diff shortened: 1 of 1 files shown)


cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -709,7 +709,7 @@ class TLMOptions(TypedDict):
         and helping catch answers that are obviously incorrect/bad for a prompt asking for a well-defined answer that LLMs should be able to handle.
         Setting this to False disables the use of self-reflection and may produce worse TLM trustworthiness scores, but will reduce costs/runtimes.
 
-        log (List[str], default = None): optionally specify additional logs or metadata to return.
+        log (List[str], default = []): optionally specify additional logs or metadata to return.
```
ulya-tkch, line 712: Wouldn't the default still be None since log is specified as NotRequired?

ulya-tkch (approved):
Change lgtm. If this works in staging this can be merged. 
Left 1 comment to address but can be merged after

cleanlab_studio/studio/trustworthy_language_model.py
huiwengoh: The defaults here are a little confusing, as they are all not actually specified in the client side code (eg. `num_candidate_responses (int, default = 1)` above is not explicitly specified but rather it is the setting we have for the `medium` preset.

I'm choosing to state `[]` as the default instead of `None` to prevent users from specifying `options={"log": None}` since that is an invalid input

PR #291, opened 2024-08-20
Feature: add bigquery upload

ryansingman:
Adds BigQuery upload to Python client

### DIFF
--- a/cleanlab_studio/studio/studio.py
+++ b/cleanlab_studio/studio/studio.py
@@ -156,6 +156,8 @@ class Studio:
         """
         Uploads a dataset, from BigQuery, to Cleanlab Studio.
 
+        Ensure that you've enabled Cleanlab Studio to access your BigQuery table, as outlined in the [BigQuery tutorial](/tutorials/notebooks/cleanlab-studio-api/bigquery_dataset).
+
         Args:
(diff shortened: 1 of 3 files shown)


cleanlab_studio/studio/studio.py
```diff
@@ -142,6 +146,40 @@ def upload_from_url(
...
+        Uploads a dataset, from BigQuery, to Cleanlab Studio.
+
+        Args:
+            bigquery_project: BigQuery project ID.
+            bigquery_dataset_id: BigQuery dataset ID.
+            bigquery_table_id: BigQuery table ID.
+            schema_overrides: Optional list of overrides you would like to make to the schema of your dataset. If not provided, all columns will be untyped. Format defined [here](/guide/concepts/datasets/#schema-updates).
+        """
+        if isinstance(schema_overrides, dict):
+            # TODO: link to documentation for schema override format
```
axl1313, line 168: nit: just noticed this also exists in the other places we have schema_overrides, but I think this TODO is no longer relevant since we have the link in the docstring

axl1313 (commented):
mostly looks good! i think there's just one place a function wasn't properly renamed

axl1313 (approved):
looks good!

PR #294, opened 2024-08-22
add bigframe upload support

axl1313:
This PR adds support for uploading from BigFrames. It also updates the BigQuery upload implementation to set up the required permissions instead of requiring the user to do that.


ryansingman (approved):
LGTM!

PR #302, opened 2024-09-04
[TLM] Enable calling get_trustworthiness_score in base preset

huiwengoh:
Note: do not merge until backend changes are in

### DIFF
--- a/cleanlab_studio/studio/trustworthy_language_model.py
+++ b/cleanlab_studio/studio/trustworthy_language_model.py
@@ -691,6 +691,7 @@ class TLMOptions(TypedDict):
     - **low:** `num_candidate_responses` = 1, `num_consistency_samples` = 4, `use_self_reflection` = True.
     - **base:** `num_candidate_responses` = 1, `num_consistency_samples` = 0, `use_self_reflection` = False. This preset is equivalent to a regular LLM call.
         When using the get_trustworthiness_score() on base preset, one self-reflection prompt will be used to generate the trustworthiness score.
+        If you explicitly set `use_self_reflection` = False for get_trustworthiness_score(), the trustworthiness score will be `None` as no evaluation is performed.
 
     By default, the TLM is set to the "medium" quality preset. The default `model` used is "gpt-4o-mini", and `max_tokens` is 512 for all quality presets.
     You can set custom values for these arguments regardless of the quality preset specified.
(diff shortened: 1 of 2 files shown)


jwmueller (approved):

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -65,7 +65,8 @@ class TLM:
             with "best" also returning more reliable trustworthiness scores than "high".
             The "medium" and "low" presets return standard LLM responses along with associated trustworthiness scores,
             with "medium" producing more reliable trustworthiness scores than low.
-            The "base" preset will not return any trustworthiness score, just a standard LLM response, and is similar to directly using your favorite LLM API.
+            The "base" preset will provide a standard LLM response and a trustworthiness score in the lowest possible latency.
```
jas2600, line 68: we should also indicating that if `"use_self_reflection": False` then it will return None since we only use the self reflection for the "base" preset scoring

huiwengoh: I added a note here instead: https://github.com/cleanlab/cleanlab-studio/pull/302/commits/582e500198c367c6a4655484cefacd759ee8d60f

jwmueller, line 68:
```suggestion
            The "base" preset will provide a standard LLM response and a trustworthiness score in the lowest possible latency/cost.
```

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -695,6 +690,8 @@ class TLMOptions(TypedDict):
     - **medium:** `num_candidate_responses` = 1, `num_consistency_samples` = 8, `use_self_reflection` = True.
     - **low:** `num_candidate_responses` = 1, `num_consistency_samples` = 4, `use_self_reflection` = True.
     - **base:** `num_candidate_responses` = 1, `num_consistency_samples` = 0, `use_self_reflection` = False. This preset is equivalent to a regular LLM call.
+        When using the get_trustworthiness_score() on base preset, one self-reflection prompt will be used to generate the trustworthiness score.
```
jwmueller, line 693:
```suggestion
        When using `get_trustworthiness_score()` on "base" preset, a cheaper self-reflection will be used to approximate the trustworthiness score.
```

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -695,6 +690,8 @@ class TLMOptions(TypedDict):
...
+        If you explicitly set `use_self_reflection` = False for get_trustworthiness_score(), the trustworthiness score will be `None` as no evaluation is performed.
```
jwmueller, line 694:
```suggestion
        If you explicitly set `use_self_reflection = False`, `get_trustworthiness_score()` will return `None` instead of a score.
```

PR #303, opened 2024-09-05
[TLM] Extend TLM testing

ulya-tkch:


tests/tlm/test_get_trustworthiness_score.py
```diff
@@ -8,7 +8,9 @@
 
 def is_trustworthiness_score(response: Any) -> bool:
     """Returns True if the response is a trustworthiness score."""
-    return isinstance(response, float)
+    return isinstance(response, float) or (
+        isinstance(response, dict) and "trustworthiness_score" in response
```
jwmueller, line 12:
```suggestion
        isinstance(response, dict) and "trustworthiness_score" in response and isinstance(response["trustworthiness_score"], float)
```

jwmueller: can you also test the floats are between 0 and 1

.github/workflows/ci-run-on-main.yml
```diff
@@ -0,0 +1,76 @@
+name: Push to Main CI
```
huiwengoh, line 1: I would rename this file to something more descriptive and has TLM in the name: `ci-tlm-main.yml` ?

ulya-tkch: I wanted to keep it general in case we add other non-tlm related tests here but I have no strong opinions

ulya-tkch:
If I add property tests to regular CI, they complete without issue: https://github.com/cleanlab/cleanlab-studio/actions/runs/10836616789/job/30070762056?pr=303

.github/workflows/ci-test-tlm.yml
```diff
@@ -42,7 +42,7 @@ jobs:
         run:  cleanlab login --key "${{ secrets.CLEANLAB_STUDIO_CI_API_KEY }}"
       - name: Run tests
         run: |
-          pytest tests/tlm/ --ignore=tests/tlm/test_properties.py --verbose
+          pytest --ignore=tests/tlm/test_properties.py --verbose
```
huiwengoh, line 45:
```suggestion
          pytest tests/tlm/ --ignore=tests/tlm/test_properties.py --verbose
```

ulya-tkch: how does this keep getting erased :o wild

cleanlab_studio/internal/settings.py
```diff
@@ -73,10 +78,13 @@ def update_version(self) -> None:
...
     def validate_version(self) -> None:
-        if semver.compare(MIN_SETTINGS_VERSION, self.version) == 1:
+        min_version = semver.VersionInfo.parse(MIN_SETTINGS_VERSION)
+        max_version = semver.VersionInfo.parse(MAX_SETTINGS_VERSION)
+        current_version = semver.VersionInfo.parse(self.version)
+        if current_version < min_version:
             # TODO add proper settings migrations
             raise SettingsError("Settings file must be migrated or re-generated.")
-        elif semver.compare(MAX_SETTINGS_VERSION, self.version) == -1:
+        elif current_version > max_version:
```
huiwengoh, line 87: Curious how this change relates to the rest of the PR?

tests/tlm/test_prompt.py
```diff
@@ -8,11 +8,13 @@
...
     """Returns True if the response is a TLMResponse."""
-    return (
+    if (
         isinstance(response, dict)
         and "response" in response
         and "trustworthiness_score" in response
-    )
+        and isinstance(response["trustworthiness_score"], float)
+    ):
+        return 0.0 <= response["trustworthiness_score"] <= 1.0
```
huiwengoh, line 17: You need a `return False` as well if this block does get executed (otherwise it returns `None`)

cleanlab_studio/internal/settings.py
ulya-tkch: It was throwing errors in the CI otherwise

tests/tlm/test_properties.py
```diff
@@ -1,3 +1,204 @@
...
-# Empy file - no code to test yet, just so CI passes
+from cleanlab_studio.internal.constants import (
+    _VALID_TLM_MODELS,
+    _VALID_TLM_QUALITY_PRESETS,
+)
+from cleanlab_studio.studio.trustworthy_language_model import TLM
+from tests.tlm.test_get_trustworthiness_score import is_trustworthiness_score
+from tests.tlm.test_prompt import is_tlm_response
+
+_VALID_TLM_MODELS = ["gpt-4o-mini", "gpt-3.5-turbo-16k", "gpt-4", "gpt-4o", "claude-3-haiku"]
```
huiwengoh, line 14: Just curious: are we making the decision to ignore the models that are not publicly documented?

huiwengoh, line 14: Also, `_VALID_TLM_MODELS` is additionally imported above, so this var is overloaded

tests/tlm/test_properties.py
```diff
@@ -1,3 +1,204 @@
...
+    """Property tests the responses of a prompt based on the options dictionary and returned responses."""
+    assert response is not None
+    assert is_tlm_response(response)
+    _test_log(response, options)
+
+
+def _test_batch_prompt_response(responses, options):
+    """Property tests the responses of a batch prompt based on the options dictionary and returned responses."""
+    assert responses is not None
+    assert isinstance(responses, list)
```
huiwengoh, line 47: We also have to test that every item in this is `is_tlm_response()` right?

huiwengoh: Ah nvm I see it [here](https://github.com/cleanlab/cleanlab-studio/blob/8f42ba3741ac1daade546bb80b352c541bb60f18/tests/tlm/test_properties.py#L135) -- would it make more sense to shift it in this function tho?

tests/tlm/conftest.py
```diff
@@ -26,7 +32,59 @@ def tlm(studio: Studio) -> TLM:
...
+    add_num_candidate_responses = np.random.choice([True, False])
+    add_num_consistency_samples = np.random.choice([True, False])
+    add_use_self_reflection = np.random.choice([True, False])
+    add_log_explanation = np.random.choice([True, False])
+    add_log_perplexity_score = np.random.choice([True, False])
+
+    if add_max_tokens:
+        options["max_tokens"] = int(np.random.randint(64, 512))
+    if add_use_self_reflection:
+        options["use_self_reflection"] = True
```
huiwengoh, line 75: I think you need the case where `use_self_reflection` is `False` as well

tests/tlm/test_properties.py
ulya-tkch: Jonas suggested to run on all models we have a high enough rate limit for

ulya-tkch: Overloaded initially for testing with a single model but good catch, I will properly handle this

tests/tlm/test_properties.py
ulya-tkch: This is different depending on if its a try_ function (where response could be None) and a regular function, that is why I kept it out of this. 

With Jay's change this test will change overall so I think we can just ignore for now

tests/tlm/conftest.py
ulya-tkch: This was giving huge typing errors (np._bool vs bool) so I removed it but good catch will add it back

tests/tlm/test_properties.py
```diff
@@ -11,7 +11,7 @@
 from tests.tlm.test_get_trustworthiness_score import is_trustworthiness_score
 from tests.tlm.test_prompt import is_tlm_response
 
-_VALID_TLM_MODELS = ["gpt-4o-mini", "gpt-3.5-turbo-16k", "gpt-4", "gpt-4o", "claude-3-haiku"]
+valid_tlm_models = [model for model in _VALID_TLM_MODELS if "sonnet" not in model]
```
huiwengoh, line 14: [nit] I'd rather explicitly state them:

```suggestion
models_to_skip = ["claude-3-sonnet", "claude-3.5-sonnet"]  # skipping due to low rate limits
valid_tlm_models = [model for model in _VALID_TLM_MODELS if model not in models_to_skip]
```

tests/tlm/test_properties.py
```diff
@@ -3,15 +3,12 @@
...
-    _VALID_TLM_MODELS,
-    _VALID_TLM_QUALITY_PRESETS,
-)
+from cleanlab_studio.internal.constants import _VALID_TLM_QUALITY_PRESETS
 from cleanlab_studio.studio.trustworthy_language_model import TLM
 from tests.tlm.test_get_trustworthiness_score import is_trustworthiness_score
 from tests.tlm.test_prompt import is_tlm_response
 
-valid_tlm_models = [model for model in _VALID_TLM_MODELS if "sonnet" not in model]
+valid_tlm_models = ["gpt-4o-mini", "gpt-3.5-turbo-16k", "gpt-4", "gpt-4o", "claude-3-haiku"]
```
huiwengoh, line 11: Sorry this is kinda nit, but I think we rather explicitly exclude that include right? Like if we add a new model, we want it to automatically be tested

ulya-tkch: Oh I think I misunderstood your review here. I thought you wanted an explicit list of IN models. Yep

huiwengoh (approved):

PR #304, opened 2024-09-06
[TLM] log error msg from try_prompt and try_get_trustworthiness_score

jas2600:
Changes:

- This is a backwards incompatible change so bumped the major version
- All the score related functions only return TLMScore objects. When it’s a successful call and no log is returned, it will be a single “trustworthiness_score” field inside the json
- Both the try_prompt() and try_get_trustworthiness_score() will no longer return None
- Changed tlm_lite in the same format as well
- In addition to the final error log output, we also raise warning to the user as well so they can be aware of

Tests:
https://drive.google.com/file/d/1v-pBme_ENiUObEnAaxdi9fmQyVR2lJyc/view?usp=drive_link

### DIFF
--- a/cleanlab_studio/internal/api/api.py
+++ b/cleanlab_studio/internal/api/api.py
@@ -115,13 +115,14 @@ async def handle_tlm_client_error_from_resp(
         try:
             res_json = await resp.json()
             error_message = res_json["error"]
+            retryable = False
         except Exception:
             error_message = "TLM query failed. Please try again and contact support@cleanlab.ai if the problem persists."
-
+            retryable = True
         if batch_index is not None:
             error_message = f"Error executing query at index {batch_index}:\n{error_message}"
 
-        raise TlmBadRequest(error_message)
+        raise TlmBadRequest(error_message, retryable)
 
 
 async def handle_tlm_api_error_from_resp(
--- a/cleanlab_studio/studio/trustworthy_language_model.py
+++ b/cleanlab_studio/studio/trustworthy_language_model.py
@@ -416,7 +416,7 @@ class TLM:
             List[TLMResponse]: list of [TLMResponse](#class-tlmresponse) objects containing the response and trustworthiness score.
                 The returned list will always have the same length as the input list.
                 In case of TLM failure on any prompt (due to timeouts or other errors),
-                the return list will include a [TLMResponse](#class-tlmresponse) with an error message and retryability information instead of the usual TLM response for that failed prompt.
+                the return list will include a [TLMResponse](#class-tlmresponse) with an error message and retryability information instead of the usual TLMResponse for that failed prompt.
                 Use this method to obtain TLM results for as many prompts as possible, while handling errors/timeouts manually.
                 If you prefer immediate notification about any errors or timeouts when processing multiple prompts,
                 use the [`prompt()`](#method-prompt) method instead.
--- a/cleanlab_studio/studio/trustworthy_language_model.py
+++ b/cleanlab_studio/studio/trustworthy_language_model.py
@@ -462,6 +462,18 @@ class TLM:
         capture_exceptions: bool = False,
         batch_index: Optional[int] = None,
     ) -> TLMResponse:
+        """
+        Private asynchronous method to get response and trustworthiness score from TLM.
+
+        Args:
+            prompt (str): prompt for the TLM
+            client_session (aiohttp.ClientSession, optional): async HTTP session to use for TLM query. Defaults to None (creates a new session).
+            timeout: timeout (in seconds) to run the prompt, defaults to None (no timeout)
+            capture_exceptions (bool): if True, the returned [TLMResponse](#class-tlmresponse) object will include error details and retry information if any errors or timeouts occur during processing.
+            batch_index: index of the prompt in the batch, used for error messages
+        Returns:
+            TLMResponse: [TLMResponse](#class-tlmresponse) object containing the response and trustworthiness score.
+        """
         response_json = await asyncio.wait_for(
             api.tlm_prompt(
                 self._api_key,
@@ -632,6 +644,18 @@ class TLM:
         capture_exceptions: bool = False,
         batch_index: Optional[int] = None,
     ) -> TLMScore:
+        """Private asynchronous method to get trustworthiness score for prompt-response pairs.
+
+        Args:
+            prompt: prompt for the TLM to evaluate
+            response: response corresponding to the input prompt
+            client_session: async HTTP session to use for TLM query. Defaults to None.
+            timeout: timeout (in seconds) to run the prompt, defaults to None (no timeout)
+            capture_exceptions (bool): if True, the returned [TLMScore](#class-tlmscore) object will include error details and retry information if any errors or timeouts occur during processing.
+            batch_index: index of the prompt in the batch, used for error messages
+        Returns:
+            [TLMScore](#class-tlmscore) objects with error messages and retryability information in place of the trustworthiness score
+        """
         response_json = await asyncio.wait_for(
             api.tlm_get_confidence_score(
                 self._api_key,
--- a/cleanlab_studio/studio/trustworthy_language_model.py
+++ b/cleanlab_studio/studio/trustworthy_language_model.py
@@ -754,7 +754,7 @@ class TLMResponse(TypedDict):
         log (dict, optional): additional logs and metadata returned from the LLM call only if the `log` key was specified in TLMOptions.
     """
 
-    response: str
+    response: Optional[str]
     trustworthiness_score: Optional[float]
     log: NotRequired[Dict[str, Any]]
 
--- a/cleanlab_studio/studio/trustworthy_language_model.py
+++ b/cleanlab_studio/studio/trustworthy_language_model.py
@@ -12,18 +12,7 @@ import asyncio
 import sys
 import warnings
 from functools import wraps
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    Callable,
-    Coroutine,
-    Dict,
-    List,
-    Optional,
-    Sequence,
-    Union,
-    cast,
-)
+from typing import Any, Callable, Coroutine, Dict, List, Optional, Sequence, Union, cast
 
 import aiohttp
 from tqdm.asyncio import tqdm_asyncio
@@ -63,6 +52,11 @@ def handle_tlm_exceptions(
     [Callable[..., Coroutine[Any, Any, Union[TLMResponse, TLMScore]]]],
     Callable[..., Coroutine[Any, Any, Union[TLMResponse, TLMScore]]],
 ]:
+    """Decorator to handle exceptions for TLM API calls.
+
+    lazydocs: ignore
+    """
+
     def decorator(
         func: Callable[..., Coroutine[Any, Any, Union[TLMResponse, TLMScore]]]
     ) -> Callable[..., Coroutine[Any, Any, Union[TLMResponse, TLMScore]]]:
--- a/cleanlab_studio/studio/trustworthy_language_model.py
+++ b/cleanlab_studio/studio/trustworthy_language_model.py
@@ -486,7 +486,7 @@ class TLM:
             response (str | Sequence[str]): existing response (or list of responses) associated with the input prompts.
                 These can be from any LLM or human-written responses.
         Returns:
-            TLMScore | List[TLMScore]: If a single prompt/response pair was passed in, method returns a TLMScore object containing the trustworthiness score and optional log dictionary keys.
+            TLMScore | List[TLMScore]: If a single prompt/response pair was passed in, method returns a [TLMScore](#class-tlmscore) object containing the trustworthiness score and optional log dictionary keys.
 
                 If a list of prompt/responses was passed in, method returns a list of TLMScore objects each containing the trustworthiness score and optional log dictionary keys for each prompt-response pair passed in.
 
--- a/cleanlab_studio/studio/trustworthy_language_model.py
+++ b/cleanlab_studio/studio/trustworthy_language_model.py
@@ -22,9 +22,9 @@ from typing_extensions import (  # for Python <3.11 with (Not)Required
 )
 
 from cleanlab_studio.errors import (
+    APITimeoutError,
     RateLimitError,
     TlmBadRequest,
-    TlmPartialSuccess,
     TlmServerError,
     ValidationError,
 )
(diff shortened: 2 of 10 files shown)


cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -438,17 +478,17 @@ def get_trustworthiness_score(
...
+    ) -> Union[TLMScore, List[TLMScore]]:
         """Computes trustworthiness score for arbitrary given prompt-response pairs.
 
         Args:
             prompt (str | Sequence[str]): prompt (or list of prompts) for the TLM to evaluate
             response (str | Sequence[str]): existing response (or list of responses) associated with the input prompts.
                 These can be from any LLM or human-written responses.
         Returns:
-            TLMScoreResponse | TLMBatchScoreResponse: If a single prompt/response pair was passed in, method returns either a float (representing the output trustworthiness score) or a TLMScore object containing both the trustworthiness score and log dictionary keys. See the documentation for [TLMScoreResponse](#type-aliases) for more details.
+            TLMScore | List[TLMScore]: If a single prompt/response pair was passed in, method returns a TLMScore object containing the trustworthiness score and optional log dictionary keys.
```
huiwengoh, line 489:
```suggestion
            TLMScore | List[TLMScore]: If a single prompt/response pair was passed in, method returns a [TLMScore](#class-tlmscore) object containing the trustworthiness score and optional log dictionary keys.
```

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -625,11 +653,58 @@ async def _get_trustworthiness_score_async(
...
+                warning_message = (
+                    f"prompt[{batch_index}] failed. Worth retrying. Error: {str(e.message)}"
+                )
+                warnings.warn(warning_message)
+
+                return TLMScore(
+                    trustworthiness_score=None,
+                    log={"error": {"message": str(e.message), "retryable": True}},
+                )
+            raise e
```
huiwengoh, line 698: This seems to be exactly duplicated with the `_prompt_async()` method, maybe consider putting this in some helper function?

cleanlab_studio/utils/tlm_lite.py
```diff
@@ -110,12 +116,20 @@ def prompt(
         """
         prompt_response = self._tlm_response.prompt(prompt)
 
-        if isinstance(prompt, Sequence) and isinstance(prompt_response, list):
-            response = [r["response"] for r in prompt_response]
+        if (
+            isinstance(prompt, Sequence)
+            and isinstance(prompt_response, list)
+            and all(r["response"] is not None for r in prompt_response)
```
huiwengoh, line 122: I don't think there would be an instance where `r["response"] is None` here right? Since we are using `tlm.prompt()` it should just fail?

cleanlab_studio/utils/tlm_lite.py
```diff
@@ -110,12 +116,20 @@ def prompt(
...
+        ):
+            response = cast(List[str], [r["response"] for r in prompt_response])
             perplexity = [r["log"]["perplexity"] for r in prompt_response]
             return self._batch_score(prompt, response, perplexity)
 
-        elif isinstance(prompt, str) and isinstance(prompt_response, dict):
+        elif (
+            isinstance(prompt, str)
+            and isinstance(prompt_response, dict)
+            and prompt_response["response"] is not None
```
huiwengoh, line 131: Same question as above re `prompt_response["response"]` being `None`

cleanlab_studio/utils/tlm_lite.py
```diff
@@ -175,8 +189,6 @@ def _score(
 
         if isinstance(score_response, dict):
             return {"response": response, **score_response}
-        elif isinstance(score_response, (float, int)):
-            return {"response": response, "trustworthiness_score": score_response}
         else:
             raise ValueError(f"score_response has invalid type {type(score_response)}")
```
huiwengoh, line 193: If `score_response` is always expected to be a `dict` then we don't need the else statement anymore

cleanlab_studio/utils/tlm_lite.py
```diff
@@ -206,12 +218,6 @@ def _batch_score(
...
             score_response = cast(List[TLMScore], score_response)
             return [{"response": res, **score} for res, score in zip(response, score_response)]
-        elif all(isinstance(score, (float, int)) for score in score_response):
-            score_response = cast(List[float], score_response)
-            return [
-                {"response": res, "trustworthiness_score": score}
-                for res, score in zip(response, score_response)
-            ]
         else:
             raise ValueError(f"score_response has invalid type")
```
huiwengoh, line 222: Same here re needing the else statement

cleanlab_studio/utils/tlm_lite.py
```diff
@@ -239,18 +245,11 @@ def _try_batch_score(
...
                 for res, score in zip(response, score_response)
             ]
-        elif all(score is None or isinstance(score, (float, int)) for score in score_response):
-            score_response = cast(List[Optional[float]], score_response)
-            return [
-                {"response": res, "trustworthiness_score": score} if score else None
-                for res, score in zip(response, score_response)
-            ]
         else:
             raise ValueError(f"score_response has invalid type")
```
huiwengoh, line 254: Same here re else statement

cleanlab_studio/studio/trustworthy_language_model.py
huiwengoh: There are probably more instances of these, would be good if we can link all of them

cleanlab_studio/internal/api/api.py
```diff
@@ -113,13 +113,14 @@ async def handle_tlm_client_error_from_resp(
         try:
             res_json = await resp.json()
             error_message = res_json["error"]
+            retryable = False
         except Exception:
             error_message = "TLM query failed. Please try again and contact support@cleanlab.ai if the problem persists."
-
+            retryable = True
```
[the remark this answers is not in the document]
jas2600: i don't have specific example but it means the response is not json or it doesn't have the "error" field. That shouldn't happen in a typical 4XX client error scenarios and it's likely a temporary error from server side so treat this type of exceptions as retriable

cleanlab_studio/utils/tlm_lite.py
jas2600: yes, i believe i added this to pass mypy checks

cleanlab_studio/utils/tlm_lite.py
jas2600: added this to pass mypy checks

cleanlab_studio/utils/tlm_lite.py
jas2600: this is also just trying to pass the mypy check, but i don't think it hurts to add the check and raise

cleanlab_studio/utils/tlm_lite.py
jas2600: this is also just trying to pass the mypy check

cleanlab_studio/utils/tlm_lite.py
jas2600: this is also just trying to pass the mypy check

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -201,7 +288,7 @@ async def _batch_get_trustworthiness_score(
             capture_exceptions (bool): if should return None in place of the response for any errors or timeout processing some inputs
```
ulya-tkch, line 288: if ``True``, the returned list will contain [TLMScore](#class-tlmscore) objects with error messages and retryability information in place of the score for any errors or timeout when processing a particular prompt from the batch.

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -305,47 +392,40 @@ def prompt(
...
-    ) -> List[Optional[TLMResponse]]:
+    ) -> List[TLMResponse]:
         """
         Gets response and trustworthiness score for any batch of prompts,
-        handling any failures (errors of timeouts) by returning None in place of the failures.
+        handling any failures (errors or timeouts) by returning error messages and retryability information in place of the failures.
 
-        The list returned will have the same length as the input list, if there are any
-        failures (errors or timeout) processing some inputs, the list will contain None in place of the response.
+        The list returned will have the same length as the input list. If there are any
```
ulya-tkch, line 408: Is this not the same sentence as what is written above?

> handling any failures (errors or timeouts) by returning error messages and retryability information in place of the failures.

I'd shorten the original description as
"Gets response and trustworthiness score for any batch of prompts handling any failures (errors or timeouts)" and then expand below how this is handled

jas2600:
@ulya-tkch for 
TLM Property Tests Results: ❌❌❌❌❌

do you know why it fails? i don't see meaningful error log from https://github.com/cleanlab/cleanlab-studio/actions/runs/10839334125/job/30079497021

ulya-tkch:
> @ulya-tkch for TLM Property Tests Results: ❌❌❌❌❌
> 
> do you know why it fails? i don't see meaningful error log from https://github.com/cleanlab/cleanlab-studio/actions/runs/10839334125/job/30079497021

Yes, we have a good theory! So the "issue_comment" github action only sees the changes that are in `main` branch. The file with the actual changes is in [this](https://github.com/cleanlab/cleanlab-studio/pull/303) pr which needs to be merged. 

I suggest running these tests locally for now or that PR is up for final review I can ping you the second it's merged into main (and hopefully this is fixed)

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -305,47 +392,39 @@ def prompt(
...
         Returns:
-            List[Optional[TLMResponse]]: list of [TLMResponse](#class-tlmresponse) objects containing the response and trustworthiness score.
+            List[TLMResponse]: list of [TLMResponse](#class-tlmresponse) objects containing the response and trustworthiness score.
                 The returned list will always have the same length as the input list.
                 In case of TLM failure on any prompt (due to timeouts or other errors),
-                the return list will contain None in place of the TLM response for that failed prompt.
-                Use this to obtain TLM results for as many prompts as possible,
-                but you might miss out on certain error messages.
-                If you prefer to be notified immediately about any errors or timeouts when running many prompts,
+                the return list will include a [TLMResponse](#class-tlmresponse) with an error message and retryability information instead of the usual TLM response for that failed prompt.
```
ulya-tkch, line 419:
```suggestion
                the return list will include a [TLMResponse](#class-tlmresponse) with an error message and retryability information instead of the usual TLMResponse for that failed prompt.
```

jas2600: updated manually as the suggestion doesn't work anymore after the new commit

ulya-tkch:
> TLM Property Tests Results: ❌❌❌❌❌

These tests seem to be failing due to OpenAI errors. @huiwengoh is investigating backend failures

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -378,50 +451,30 @@ async def prompt_async(
...
     async def _prompt_async(
         self,
         prompt: str,
         client_session: Optional[aiohttp.ClientSession] = None,
         timeout: Optional[float] = None,
         capture_exceptions: bool = False,
         batch_index: Optional[int] = None,
-    ) -> Optional[TLMResponse]:
-        """
-        Private asynchronous method to get response and trustworthiness score from TLM.
```
ulya-tkch, line 395: Seems the docstring here got deleted

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -588,49 +631,29 @@ async def _get_trustworthiness_score_async(
         timeout: Optional[float] = None,
         capture_exceptions: bool = False,
         batch_index: Optional[int] = None,
-    ) -> Optional[TLMScoreResponse]:
-        """Private asynchronous method to get trustworthiness score for prompt-response pairs.
-
-        Args:
```
ulya-tkch, line 594: Same here with args getting deleted

ulya-tkch (commented):
Some nit comments

ulya-tkch (approved):
Changes look good to me, make sure you notfy community at the same time as you merge this

cleanlab_studio/studio/trustworthy_language_model.py
```diff
@@ -46,6 +46,87 @@
 from cleanlab_studio.internal.types import TLMQualityPreset
 
 
+def handle_tlm_exceptions(
```
huiwengoh, line 49: Would it make more sense to move this (+ the `_handle_exception` function) to some utils file?

huiwengoh: Ah okay, let's just move these functions to the bottom of the file then? (I think it'll be better to have the main TLM class at the top)

jas2600: have also tried that but i believe decorators needs to be defined before the usage otherwise it will complain for "not defined"

huiwengoh:
Overall looking great! I think there is one last case that is not handled properly -- timeouts:

```python
tlm = studio.TLM(timeout=0.1) 
tlm.try_prompt(["what is 1+1"])
```

returns this:
```python
Querying TLM...   0%||
/Users/huiwengoh/Documents/Cleanlab/Code/branches/tlm/cleanlab_studio/studio/trustworthy_language_model.py:110: UserWarning: prompt[0] failed. Worth retrying. Error: 
  warnings.warn(warning_message)
Querying TLM... 100%|███████████|
[{'response': None,
  'trustworthiness_score': None,
  'log': {'error': {'message': '', 'retryable': True}}}]
```

jas2600:
> Overall looking great! I think there is one last case that is not handled properly -- timeouts:
> 
> ```python
> tlm = studio.TLM(timeout=0.1) 
> tlm.try_prompt(["what is 1+1"])
> ```
> 
> returns this:
> 
> ```python
> Querying TLM...   0%||
> /Users/huiwengoh/Documents/Cleanlab/Code/branches/tlm/cleanlab_studio/studio/trustworthy_language_model.py:110: UserWarning: prompt[0] failed. Worth retrying. Error: 
>   warnings.warn(warning_message)
> Querying TLM... 100%|███████████|
> [{'response': None,
>   'trustworthiness_score': None,
>   'log': {'error': {'message': '', 'retryable': True}}}]
> ```

great catch! updated.

cleanlab_studio/version.py
```diff
@@ -1,7 +1,7 @@
 # Note to developers:
 # Consider if backend's MIN_CLI_VERSION needs updating when pushing any changes to this file.
 
-__version__ = "2.4.4"
+__version__ = "2.5.4"
```
huiwengoh, line 4:
```suggestion
__version__ = "2.5.0"
```

huiwengoh (approved):
LGTM - great work!

PR #309, opened 2024-09-09
[TLM] Add base URL as optional config to CI

ulya-tkch:


.github/workflows/ci-tlm.yml
```diff
@@ -0,0 +1,75 @@
...
+    strategy:
+      matrix:
+        os:
+          - ubuntu-latest
+          - macos-latest
+          - windows-latest
+        python:
+          - "3.8"
+          - "3.9"
+          - "3.10"
```
huiwengoh, line 21: Is there a reason why we're only testing on these 3 versions of python here, and then python 3.11 for the property tests below?

ulya-tkch: Legacy, the tests were never ran on python 3.11. I can extend the prior normal tests to python 3.11 as well we would increase the number of calls from 180 -> 240. I did not initially do that to avoid the call increase and since the property tests are testing the usage already.

Property tests should only run on one python version since it is ~200-400 calls to the TLM. If we did this for all versions/all OS it would be too many calls

ulya-tkch: Yeah I added the other versions. Mostly to do with throwing errors/concurency timing but its not too many extra calls <200 for a more thorough test

huiwengoh: Are these mostly running on `gpt-4o-mini`? If so then yeah ~200 calls is not an issue at all

ulya-tkch: Yep!

huiwengoh:
This is not directly related to your changes in this PR, but I'm noticing that the CI gets run twice due to both "push" and "PR" actions triggering.

image


I believe if you update [this line](https://github.com/cleanlab/cleanlab-studio/blob/main/.github/workflows/ci.yml#L3) to the following it should prevent the duplicated runs:

```yml
on:
  push:
    branches: [main]
  pull_request:
    branches: [main]
```

huiwengoh (approved):

ulya-tkch:
I don't think your suggestion is equivalent, ci should run on PRs opened against any branch (not just main) .
```python
on:
  push:         # to any branch
  pull_request:  # to any branch
  ```
  should be more equivalent

PR #310, opened 2024-09-10
[TLM] handle ssl exception

jas2600:
Screenshot 2024-09-10 at 3 56 26 PM

### DIFF
--- a/cleanlab_studio/internal/api/api.py
+++ b/cleanlab_studio/internal/api/api.py
@@ -3,7 +3,9 @@ from __future__ import annotations
 import asyncio
 import io
 import os
+import ssl
 import time
+import warnings
 from io import StringIO
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union, cast
 
@@ -917,6 +919,12 @@ def tlm_retry(func: Callable[..., Any]) -> Callable[..., Any]:
             await asyncio.sleep(sleep_time)
             try:
                 return await func(*args, **kwargs)
+            except ssl.SSLCertVerificationError as e:
+                warnings.warn(f"SSL Certificate Verification Error: {str(e)}")
+                warnings.warn(
+                    "Please ensure that your SSL certificates are up to date. If you installed python via python pkg installer, please make sure to execute the Install Certificates.command in the python installation directory."
+                )
+                raise
             except aiohttp.client_exceptions.ClientConnectorError as e:
                 # note: we don't increment num_try here, because we don't want connection errors to count against the total number of retries
                 sleep_time = 2**num_try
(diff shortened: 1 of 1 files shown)


huiwengoh:
Do you mind showing a quick screenshot of when this warning/error is thrown?

jas2600:
> Do you mind showing a quick screenshot of when this warning/error is thrown? (can't test locally)

@huiwengoh added in the description

huiwengoh:
The really long error is raised twice (once in `str(e)` and another time when you actually `raise`). I don't think it has to be repeated? Otherwise our warning message gets drowned out

jas2600:
> The really long error is raised twice (once in `str(e)` and another time when you actually `raise`). I don't think it has to be repeated? Otherwise our warning message gets drowned out

removed the first warning but what takes long is actually not the `str(e)', it's the `The above exception was the direct cause of the following exception:` part.

cleanlab_studio/internal/api/api.py
```diff
@@ -917,6 +919,11 @@ async def wrapper(*args: Any, **kwargs: Any) -> Any:
             await asyncio.sleep(sleep_time)
             try:
                 return await func(*args, **kwargs)
+            except ssl.SSLCertVerificationError as e:
+                warnings.warn(
+                    "Please ensure that your SSL certificates are up to date. If you installed python via python pkg installer, please make sure to execute the Install Certificates.command in the python installation directory."
```
ulya-tkch, line 924:
```suggestion
                    "Please ensure that your SSL certificates are up to date. If you installed python via python pkg installer, please make sure to execute 'Install Certificates.command' in the python installation directory."
```

[very nit] with the weird space in the filename, probably want to make it more clear Install is part of the name in case someone searches for it

ulya-tkch:
We can use some color package like termcolor to improve readability for the actual warning if we find that people are missing the solution due to the large output. Possibly add a few new lines before/after warning text for more readability as well?

jas2600:
> We can use some color package like termcolor to improve readability for the actual warning if we find that people are missing the solution due to the large output. Possibly add a few new lines before/after warning text for more readability as well?

@ulya-tkch do you mind commit? i have no experience with the package. i can test and show the error message (given you can not reproduce)

ulya-tkch:
[nit] Should we still raise APIError? 
`APIError(f"TLM could not be run due to SSLCertVerificationError.", -1)`

ulya-tkch (approved):
For now this LGTM!

If this error is thrown non-deterministically due to network issues we might have to add some sort of retry before raising but I don't expect that to be the case

huiwengoh (approved):

jas2600:
> [nit] Should we still raise APIError? `APIError(f"TLM could not be run due to SSLCertVerificationError.", -1)`

don't think this error belong to APIError, which should be error from API server. Ideally we should probably define a new type of exception and shorten the long stack trace.

agreed that the error message could be improved, but given this happens rarely and we are actually not sure about whether all the hanging is because of this SSL issue so I will merge as the current format for now. The top priority for us now should be release this out so user won't get infinite hanging, and we can improve the format later if it turns out users getting this error message quite often

PR #320, opened 2024-09-13
[TLM CI] Add print for options dict + no options tlm tests

ulya-tkch:


tests/tlm/test_properties.py
```diff
@@ -36,59 +43,144 @@ def _test_log_batch(responses: Dict[str, Any], options: Dict[str, Any]) -> None:
...
+    options,
+    allow_none_response=False,
+    allow_null_trustworthiness_score=False,
+) -> bool:
+    """Returns true if trustworthiness score is valid based on properties for get_trustworthiness_score() functionality."""
+    if "log" in options:
+        assert isinstance(response, dict)
+    else:
+        assert isinstance(response, float)
+    _test_log(response, options)
```
huiwengoh, line 82: [nit] `_test_log(response, options)` can be in the if statement right?

tests/tlm/test_properties.py
```diff
@@ -36,59 +43,144 @@ def _test_log_batch(responses: Dict[str, Any], options: Dict[str, Any]) -> None:
...
 
+def _is_valid_prompt_response(
+    response,
+    options,
+    allow_none_response=False,
+    allow_null_trustworthiness_score=False,
+) -> bool:
+    """Returns true if prompt response is valid based on properties for prompt() functionality."""
+    _test_log(response, options)
+    if {"use_self_reflection", "quality_preset", "num_consistency_samples"}.issubset(options) and (
```
huiwengoh, line 54: [nit] `quality_preset` doesn't have to be checked I think

huiwengoh (approved):
Very nit comments, otherwise LGTM

tests/tlm/conftest.py
```diff
@@ -47,6 +47,10 @@ def tlm_dict(studio: Studio) -> Dict[str, Any]:
             tlm_dict[quality_preset][model]["tlm"] = studio.TLM(
                 quality_preset=quality_preset, options=options
             )
+            tlm_dict[quality_preset][model]["tlm_no_options"] = studio.TLM(
```
huiwengoh, line 50: Can you add a comment/docstring here about why we are instantiating 2 TLM objects here? Something about wanting to test both the random options + default options

jas2600 (approved):
Thanks for adding the print and the additional tests!

PR #324, opened 2024-09-19
[TLM] Update max_tokens for TLM to match provider limits

ulya-tkch:


jas2600:
we should also change the unit test fixture here https://github.com/cleanlab/cleanlab-studio/blob/90fb5cfb7311aca5300757d711eda97a66a6534f/tests/tlm/conftest.py#L90

if 4096 is too large for claude models for now, we can keep that 512 if the model is claude and change to 4096 if the model is gpt

jas2600 (approved):
changes LGTM. let's verify tests can pass after the backend changes deployed before merge.

PR #325, opened 2024-09-20
[TLM] Add custom eval as part of TLMOptions

huiwengoh:
TODO: documentation

PR #326, opened 2024-09-24
[TLM] Add TLMCalibrated class 

huiwengoh:
Building off the custom eval criteria in https://github.com/cleanlab/cleanlab-studio/pull/325

Sample workflow:

```python
prompts = ["what is 1+1", "what is 1+2", "what is 1+3"]
responses = ["2", "hm let me think, i think the answer is 3? am i right?", "5"]
ratings = [5, 3, 1]

custom_eval_criteria_option = {"custom_eval_criteria": [
    {"name": "Conciseness", "criteria": "Determine if the output is concise."}
]}

tlm = studio.TLM(options=custom_eval_criteria_option)
scores = tlm.get_trustworthiness_score(prompts, responses)

tlm_calibrated = studio.TLMCalibrated(options=custom_eval_criteria_option)
tlm_calibrated.fit(scores, ratings)

tlm_calibrated.get_trustworthiness_score("what is 2+2", "5")
>> {'trustworthiness_score': 0.050770278656628066,
 'log': {'custom_eval_criteria': [{'name': 'Conciseness', 'score': 0.8}]},
 'calibrated_score': 0.365}

tlm_calibrated.get_trustworthiness_score(["what is 2+2", "what is 2+3"], ["i think it is 4? maybe 3", "5"])
>> [{'trustworthiness_score': 0.6060140420912912,
  'log': {'custom_eval_criteria': [{'name': 'Conciseness',
     'score': 0.2641642614722914}]},
  'calibrated_score': 0.51},
 {'trustworthiness_score': 0.9967050946426079,
  'log': {'custom_eval_criteria': [{'name': 'Conciseness',
     'score': 0.9999999999999998}]},
  'calibrated_score': 0.79}]
```

### DIFF
--- a/cleanlab_studio/utils/tlm_calibrated.py
+++ b/cleanlab_studio/utils/tlm_calibrated.py
@@ -1,20 +1,21 @@
 """
-TLM Calibrated is a variant of the Trustworthy Language Model (TLM) that facilitates the calibration of trustworthiness scores 
+TLM Calibrated is a variant of the Trustworthy Language Model (TLM) that facilitates the calibration of trustworthiness scores
 using existing ratings for prompt-response pairs, which allows for better alignment of the TLM scores in specialized-use cases.
 
-**This module is not meant to be imported and used directly.** 
-Instead, use [`Studio.TLMCalibrated()`](/reference/python/studio/#method-tlmcalibrated) to instantiate a [TLMCalibrated](#class-tlmcalibrated) object, 
+**This module is not meant to be imported and used directly.**
+Instead, use [`Studio.TLMCalibrated()`](/reference/python/studio/#method-tlmcalibrated) to instantiate a [TLMCalibrated](#class-tlmcalibrated) object,
 and then you can use the methods like [`get_trustworthiness_score()`](#method-get_trustworthiness_score) documented on this page.
 """
 
 from __future__ import annotations
-from typing import Optional, Sequence, Union, List
+
+from typing import List, Optional, Sequence, Union, cast
 
 import numpy as np
 import numpy.typing as npt
 import pandas as pd
 
-from cleanlab_studio.errors import ValidationError, TlmNotCalibratedError
+from cleanlab_studio.errors import TlmNotCalibratedError, ValidationError
 from cleanlab_studio.internal.types import TLMQualityPreset
 from cleanlab_studio.studio.trustworthy_language_model import TLM, TLMOptions, TLMScore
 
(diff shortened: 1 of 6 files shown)


cleanlab_studio/utils/tlm_calibrated.py
```diff
@@ -0,0 +1,173 @@
...
+        tlm_scores_df = pd.DataFrame(tlm_scores)
+
+        extracted_scores = self._extract_tlm_scores(tlm_scores_df)
+
+        try:
+            tlm_scores_df["calibrated_score"] = self._rf_model.predict(extracted_scores)
+        except NotFittedError:
+            raise TlmNotCalibratedError(
+                "TLMCalibrated has to be calibrated before scoring new data, use the .fit() method to calibrate the model."
+            )
```
jas2600, line 125: is there a way to check whether the model is already trained before making the `.predict` call so we can check and raise the error earlier? The current implementation wastes the above actual calls to the TLM server to generate those scores

cleanlab_studio/utils/tlm_calibrated.py
```diff
@@ -0,0 +1,173 @@
...
+
+        # using pandas so that NaN values are handled correctly
+        ratings_series = pd.Series(ratings)
+        ratings_normalized = (ratings_series - ratings_series.min()) / (
+            ratings_series.max() - ratings_series.min()
+        )
+
+        self._rf_model.fit(extracted_scores, ratings_normalized.values)
+
+    def get_trustworthiness_score(
```
jas2600, line 100: could be in followup PRs since i believe we want to release this asap, but I think we should also support the `prompt` function for this new class

cleanlab_studio/utils/tlm_calibrated.py
huiwengoh: Ah good call, there actually is - moved it

setup.py
```diff
@@ -63,6 +63,7 @@
         "openpyxl>=3.0.0,!=3.1.0",
         "validators>=0.20.0",
         "matplotlib>=3.4.0",
+        "scikit-learn",
```
jwmueller, line 66: let's not have this as a hard dependency of the package and instead do it as a lazy try/except import inside only the tlm_calibrated file / functions where sklearn is needed

jwmueller: in the future would be good to call out in PR description if you are introducing a new dependency to the client, that is a major change

cleanlab_studio/utils/tlm_calibrated.py
jas2600: created https://www.notion.so/cleanlab/support-prompt-and-try_-functions-for-TLMCalibrated-10cc7fee85be8070b2c9ff018df04d18

jas2600:
for unit test this new class and the custom eval feature, created ticket https://www.notion.so/cleanlab/Add-unit-tests-in-cleanlab-studio-for-TLMCalibrated-and-the-custom-eval-feature-c01d32f159c7473a878a998b37f52764

jas2600 (approved):

PR #330, opened 2024-10-04
[TLM] Improve retry logic

huiwengoh:
Add a max number of retried for connection errors so that the TLM would not hang indefinitely.

### DIFF
--- a/cleanlab_studio/internal/api/api.py
+++ b/cleanlab_studio/internal/api/api.py
@@ -910,13 +910,19 @@ def tlm_retry(func: Callable[..., Any]) -> Callable[..., Any]:
 
     async def wrapper(*args: Any, **kwargs: Any) -> Any:
         # total number of tries = number of retries + original try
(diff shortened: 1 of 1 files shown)


cleanlab_studio/internal/api/api.py
```diff
@@ -926,20 +932,34 @@ async def wrapper(*args: Any, **kwargs: Any) -> Any:
                 )
                 raise
             except aiohttp.client_exceptions.ClientConnectorError as e:
-                # note: we don't increment num_try here, because we don't want connection errors to count against the total number of retries
-                sleep_time = 2**num_try
+                if num_connection_error_retry == (max_connection_error_retries // 2):
+                    warnings.warn(
+                        f"Connection error after {max_connection_error_retries // 2} retries. Retrying..."
```
jas2600, line 937: nit: use `num_connection_error_retry` directly instead of `max_connection_error_retries // 2`?

jas2600 (approved):

PR #335, opened 2024-10-24
Added refactored client side data enrichment code to match backend AP…

mturk24:
Refactored client side code based on this task: https://www.notion.so/cleanlab/make-data-enrichment-client-side-API-match-backend-API-105c7fee85be8097b54dfb121b7dba4e

Goal: make Dynamic API match the Static API. 

- in all facets: cohesive naming of methods, argument types, regular expression libraries, etc.
- This way user can use backend API to prototype Enrichment jobs and run them over a big static dataset, but then use client-API when they need to run the same logic in real-time over streaming data one at a time.
- For any packages we need to import client-side, make these lazy optional imports, so cleanlab-studio package still works without those installed.

Still doing some testing (unit test runs) but will request review on overall structure.

### DIFF
--- a/cleanlab_studio/utils/data_enrichment/enrich.py
+++ b/cleanlab_studio/utils/data_enrichment/enrich.py
@@ -1,161 +1,171 @@
-from typing import Any, List, Optional, Tuple, Union
-import pandas as pd
+from __future__ import annotations
+from typing import Any, Dict, List, Optional, Union, Tuple, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    import pandas as pd
+
+from cleanlab_studio.studio.enrichment import (
+    EnrichmentProject, 
+    EnrichmentOptions, 
+    EnrichmentResults
+)
+import re
+
 from cleanlab_studio.internal.enrichment_utils import (
-    extract_df_subset,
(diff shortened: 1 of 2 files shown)


mturk24:
Example flow of running this
```
from cleanlab_studio import Studio
from cleanlab_studio.client.enrichment import ClientEnrichmentProject
import pandas as pd

df = pd.DataFrame({'text': ['Great product', 'Bad service']})

dataset_id = Studio.upload_dataset(df)
de = Studio.create_enrichment_project(dataset_id)

options = {
    'prompt': 'What is the sentiment of this text: ${text}',
    'constrain_outputs': ['positive', 'negative'],
    'optimize_prompt': True
}

de.preview(options, 'sentiment')

results = de.run(options, 'sentiment')

# Convert to client project for online inference
client_de = ClientEnrichmentProject(api_key=de._api_key, id=de.id, name=de.name)

# Now switch to client-side for streaming
new_df = pd.DataFrame({'text': ['Amazing quality', 'Terrible experience']})

# Use exact same options for online inference
streaming_results = client_de.online_inference(new_df, options, 'sentiment')
# Returns same format as de.run()
```

PR #336, opened 2024-10-28
Refactored static and dynamic enrichment APIs

mturk24:
Refactored client side code based on this task: https://www.notion.so/cleanlab/make-data-enrichment-client-side-API-match-backend-API-105c7fee85be8097b54dfb121b7dba4e

Goal: make Dynamic API match the Static API.

in all facets: cohesive naming of methods, argument types, regular expression libraries, etc.
This way user can use backend API to prototype Enrichment jobs and run them over a big static dataset, but then use client-API when they need to run the same logic in real-time over streaming data one at a time.
For any packages we need to import client-side, make these lazy optional imports, so cleanlab-studio package still works without those installed.

Still doing some testing (unit test runs) but will request review on overall structure.

### DIFF
--- a/cleanlab_studio/studio/enrichment.py
+++ b/cleanlab_studio/studio/enrichment.py
@@ -17,6 +17,7 @@ import matplotlib.pyplot as plt
 import pandas as pd
 from tqdm import tqdm
 from typing_extensions import NotRequired
+from functools import lru_cache
 
 from cleanlab_studio.errors import EnrichmentProjectError
 from cleanlab_studio.internal.api import api
@@ -47,6 +48,12 @@ def _response_timestamp_to_datetime(timestamp_string: str) -> datetime:
     """
     response_timestamp_format_str = "%a, %d %b %Y %H:%M:%S %Z"
     return datetime.strptime(timestamp_string, response_timestamp_format_str)
+
+
+@lru_cache(maxsize=None)
+def _get_online_inference():
+    from cleanlab_studio.utils.data_enrichment.enrich import online_inference
+    return online_inference
 
 
 class EnrichmentProject:
@@ -399,6 +406,27 @@ class EnrichmentProject:
         latest_job = self._get_latest_job()
         return api.resume_enrichment_job(api_key=self._api_key, job_id=latest_job["id"])
 
+    def online_inference(
+        self,
+        data: Union[pd.DataFrame, List[dict]],
+        options: EnrichmentOptions,
+        new_column_name: str,
+    ) -> Dict[str, Any]:
+        """
+        Enrich data in real-time using the same logic as the run() method, but client-side.
+
+        Args:
+            data (Union[pd.DataFrame, List[dict]]): The dataset to enrich.
+            options (EnrichmentOptions): Options for enriching the dataset.
+            new_column_name (str): The name of the new column to store the results.
+
+        Returns:
+            Dict[str, Any]: A dictionary containing information about the enrichment job and the enriched dataset.
+        """
+        online_inference = _get_online_inference()
+        job_info = online_inference(data, options, new_column_name, None)
+        return job_info
+
 
 class EnrichmentJob(TypedDict):
     """Represents an Enrichment Job instance.
@@ -611,3 +639,7 @@ def _handle_replacements_and_extraction_pattern(
         else:
             raise ValueError(REGEX_PARAMETER_ERROR_MESSAGE)
     return extraction_pattern, replacements
+
+
+
+
--- a/cleanlab_studio/utils/data_enrichment/enrich.py
+++ b/cleanlab_studio/utils/data_enrichment/enrich.py
@@ -1,5 +1,5 @@
-from typing import Any, List, Optional, Tuple, Union
-import pandas as pd
+from typing import Any, List, Optional, Tuple, Union, Dict
+from functools import lru_cache
 from cleanlab_studio.internal.enrichment_utils import (
     extract_df_subset,
     get_prompt_outputs,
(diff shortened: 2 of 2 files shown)


cleanlab_studio/utils/data_enrichment/enrich.py
```diff
@@ -1,91 +1,71 @@
...
-        `trustworthiness` column = trustworthiness of the prompt responses (which ignore the data mutations).
-        **Note**: If you specified the `regex` or `constrain_outputs` arguments, some additional transformations may be applied to raw LLM outputs to produce the returned values. In these cases, an additional `log` column will be added to the returned DataFrame that records the raw LLM outputs (feel free to disregard these).
+        Dict[str, Any]: A dictionary containing information about the enrichment job and the enriched dataset.
     """
-    if subset_indices:
-        df = extract_df_subset(data, subset_indices)
-    else:
-        df = data.copy()
+    pd = _get_pandas()
+    tqdm = _get_tqdm()
```
huiwengoh, line 45: Is there a reason these are not just being imported at the top of the file?

cleanlab_studio/studio/enrichment.py
```diff
@@ -399,6 +409,27 @@ def resume(self) -> JSONDict:
...
+
+        Args:
+            data (Union[pd.DataFrame, List[dict]]): The dataset to enrich.
+            options (EnrichmentOptions): Options for enriching the dataset.
+            new_column_name (str): The name of the new column to store the results.
+
+        Returns:
+            Dict[str, Any]: A dictionary containing information about the enrichment job and the enriched dataset.
+        """
+        run_online = _get_run_online()
```
huiwengoh, line 429: Why is this not just imported at the top of the file?

cleanlab_studio/utils/data_enrichment/enrich.py
```diff
@@ -1,91 +1,71 @@
...
     get_constrain_outputs_match,
     get_optimized_prompt,
-    Replacement,
 )
+from cleanlab_studio.studio.enrichment import EnrichmentOptions
 
-from cleanlab_studio.studio.studio import Studio
-
-
-def enrich_data(
```
huiwengoh, line 15: Are we deprecating this `enrich_data` function?

cleanlab_studio/studio/enrichment.py
```diff
@@ -399,6 +409,27 @@ def resume(self) -> JSONDict:
...
+        job_info = run_online(data, options, new_column_name, self._api_key)
```
huiwengoh, line 430: I don't think passing in `self._api_key` because `run_online` expects a `Studio` object?

cleanlab_studio/studio/enrichment.py
```diff
@@ -399,6 +409,27 @@ def resume(self) -> JSONDict:
...
+        data: Union[pd.DataFrame, List[dict]],
+        options: EnrichmentOptions,
+        new_column_name: str,
+    ) -> Dict[str, Any]:
+        """
+        Enrich data in real-time using the same logic as the run() method, but client-side.
+
+        Args:
+            data (Union[pd.DataFrame, List[dict]]): The dataset to enrich.
+            options (EnrichmentOptions): Options for enriching the dataset.
```
huiwengoh, line 423: Link to `EnrichmentOptions` docstring

cleanlab_studio/utils/data_enrichment/enrich.py
```diff
@@ -94,41 +74,78 @@ def enrich_data(
...
+        "num_rows": len(enriched_df),
+        "processed_rows": len(enriched_df),
+        "average_trustworthiness_score": enriched_df[f"{column_name_prefix}trustworthiness"].mean(),
+        "results": enriched_df,
+    }
+
+    return job_info
+
+
+def _validate_enrichment_options(options: EnrichmentOptions) -> None:
```
huiwengoh, line 115: Can you clarify why there is a separate `_validate_enrichment_options` defined here rather than using the validation function in `run()` [here](https://github.com/cleanlab/cleanlab-studio/blob/refactor-dynamic-static-enrichment-apis/cleanlab_studio/studio/enrichment.py#L190)?

cleanlab_studio/utils/data_enrichment/enrich.py
```diff
@@ -94,41 +74,78 @@ def enrich_data(
...
+        if isinstance(regex, list) and not all(isinstance(item, tuple) for item in regex):
+            raise TypeError("All items in 'regex' list must be tuples.")
 
 
 def process_regex(
-    column_data: Union[pd.Series, List[str]],
-    regex: Union[str, Replacement, List[Replacement]],
-) -> Union[pd.Series, List[str]]:
+    column_data: Union["pd.Series", List[str]],
+    regex: Union[str, Tuple[str, str], List[Tuple[str, str]]],
```
huiwengoh, line 147: [nit] `Replacement` is a type alias for the `Tuple[str, str]` type (ref [here](https://github.com/cleanlab/cleanlab-studio/blob/refactor-dynamic-static-enrichment-apis/cleanlab_studio/internal/enrichment_utils.py#L12)), not entirely sure why you made this change?

cleanlab_studio/utils/data_enrichment/enrich.py
```diff
@@ -1,91 +1,71 @@
...
-    new_column_name: str = "metadata",
-    disable_warnings: bool = False,
-    **kwargs: Any,
-) -> pd.DataFrame:
+
+@lru_cache(maxsize=None)
+def _get_pandas():
+    import pandas as pd
+
+    return pd
```
jwmueller, line 16: what is going on here?

pandas is already a dependency of this package, there should be no special logic to lazy-import it

https://github.com/cleanlab/cleanlab-studio/blob/c2a3013fbb1b2a77d3182a7894367b357a5b9c35/setup.py#L52

cleanlab_studio/utils/data_enrichment/enrich.py
```diff
@@ -1,91 +1,71 @@
...
+
+
+@lru_cache(maxsize=None)
+def _get_tqdm():
+    from tqdm import tqdm
+
+    return tqdm
+
```
jwmueller, line 24: what is going on here?

tqdm is already a dependency of this package, there should be no special logic to lazy import it

https://github.com/cleanlab/cleanlab-studio/blob/c2a3013fbb1b2a77d3182a7894367b357a5b9c35/setup.py#L57

cleanlab_studio/studio/enrichment.py
```diff
@@ -49,6 +50,13 @@ def _response_timestamp_to_datetime(timestamp_string: str) -> datetime:
     return datetime.strptime(timestamp_string, response_timestamp_format_str)
 
 
+@lru_cache(maxsize=None)
+def _get_run_online():
+    from cleanlab_studio.utils.data_enrichment.enrich import run_online
+
+    return run_online
+
```
jwmueller, line 58: get rid of this and do a standard import, unsure why you are using some an odd approach

cleanlab_studio/studio/enrichment.py
```diff
@@ -399,6 +409,27 @@ def resume(self) -> JSONDict:
...
         return api.resume_enrichment_job(api_key=self._api_key, job_id=latest_job["id"])
 
+    def run_online(
+        self,
+        data: Union[pd.DataFrame, List[dict]],
+        options: EnrichmentOptions,
+        new_column_name: str,
+    ) -> Dict[str, Any]:
+        """
+        Enrich data in real-time using the same logic as the run() method, but client-side.
```
jwmueller, line 419:
```suggestion
        Enrich data in real-time using the same logic as the ``run()`` method. This runs client-side on your local machine, processing one record at a time. 
```

cleanlab_studio/studio/enrichment.py
```diff
@@ -17,6 +17,7 @@
 import pandas as pd
```
jwmueller, line 17: please update the PR description with:

1. User code if they just want to do some real-time data enrichment quickly.

2. User code if they want to first run data enrichment project over a big static dataset, and then later want to run some real-time data enrichment over additional data.

PR #340, opened 2024-12-18
TLM with constrain_outputs

jas2600:


cleanlab_studio/internal/tlm/validation.py
```diff
@@ -17,6 +17,36 @@
...
+    constrain_outputs = kwargs.get("constrain_outputs")
+    if constrain_outputs is not None:
+        if isinstance(prompt, str):
+            if not isinstance(constrain_outputs, list):
+                raise ValidationError("constrain_outputs must be a list of strings")
+        elif isinstance(prompt, Sequence):
+            if not isinstance(constrain_outputs, list) or not all(
+                isinstance(co, list) for co in constrain_outputs
+            ):
+                raise ValidationError("constrain_outputs must be a list of lists of strings")
```
huiwengoh, line 43: Did you need to verify that each item in the list is indeed a string? (seems like a lot to iterate through, but not sure if there will be weird failures later on if unverified)

cleanlab_studio/internal/tlm/validation.py
```diff
@@ -17,6 +17,36 @@
 )
 
 
+def validate_tlm_prompt_kwargs_constrain_outputs(
```
huiwengoh, line 20: [nit] this function could probably just be `validate_tlm_prompt_kwargs`? (in the case where we add more kwargs in the future)

cleanlab_studio/internal/tlm/validation.py
```diff
@@ -17,6 +17,36 @@
...
+    prompt: Union[str, Sequence[str]],
+    **kwargs: Any,
+) -> Union[Optional[List[str]], Optional[List[Optional[List[str]]]]]:
+    # validate kwargs - only allow constrain_outputs
+    supported_kwargs = {"constrain_outputs"}
```
huiwengoh, line 25: [nit] define this as constant here: https://github.com/cleanlab/cleanlab-studio/blob/main/cleanlab_studio/internal/constants.py ?

cleanlab_studio/internal/tlm/validation.py
jas2600: the backend will also verify so currently if the element is not a string it will get 400. But good point! It's better to verify here as well to raise error earlier. The for loop should be cheap so should be fine in term of compute cost.

cleanlab_studio/internal/tlm/validation.py
jas2600: i purposely made this specific for constrain_outputs since as you can see that the validation logic for each kwargs is fairly complicated so better to have separate validation function for each of them.

huiwengoh (approved):
lgtm!

PR #344, opened 2025-01-10
[TLM] Add more bedrock models

huiwengoh:


tests/tlm/constants.py
```diff
@@ -16,13 +16,25 @@
 CHARACTERS_PER_TOKEN: int = 4
 
 # Property tests for TLM
-excluded_tlm_models: List[str] = ["claude-3-sonnet", "claude-3.5-sonnet", "o1-preview"]
+excluded_tlm_models: List[str] = [
```
jas2600, line 19: I forgot, but what is our rules to determine whether to include models for testing?

huiwengoh: I don't know if there were concrete rules, I'm just excluding the expensive/low rate limit models

jas2600 (approved):

Issue #347, opened 2025-01-21
Dependency Conflict: Click package version incompatibility between cleanlab-studio and langgraph-cli in LangGraph Studio


sharkfabri:
### Description
When attempting to install the requirements, there is a package dependency conflict between `cleanlab-studio` and `langgraph-cli` regarding the `Click` package version requirements.

### Current Behavior
The installation fails due to conflicting dependencies:
- `cleanlab-studio 2.5.13` requires `Click >=8.1.0, <=8.1.3`
- `langgraph-cli` (all versions from 0.1.0 to 0.1.68) requires `click >=8.1.7, <9.0.0`

These requirements cannot be satisfied simultaneously as there is no version of Click that meets both constraints.

### Error Message
`ERROR: Cannot install -r requirements.txt (line 1) and -r requirements.txt (line 10) because these package versions have conflicting dependencies.`

### Steps to Reproduce
1. Try to install both `cleanlab-studio` and `langgraph-cli` in the same environment
2. The package manager fails to resolve the dependency conflict

### Possible Solutions
Update `cleanlab-studio` to support newer versions of Click (>=8.1.7)

### Additional Notes
This is a blocking issue that prevents the installation and usage of both packages together in the same environment.

PR #358, opened 2025-02-03
Allow Click 8.1.7+ for langchain etc compatibility

srowen:
Closes https://github.com/cleanlab/cleanlab-studio/issues/347
Allowing Click 8.1.7 and later fixes compatibility with packages like langchain that require later Click 8.1.x versions.


huiwengoh (approved):
Thanks for the update!

PR #360, opened 2025-02-12
[TLM] scoring add constrain outputs

jas2600:


cleanlab_studio/internal/constants.py
```diff
@@ -36,6 +36,6 @@
 TLM_SIMILARITY_MEASURES: Set[str] = {"semantic", "string"}
 TLM_REASONING_EFFORT_VALUES: Set[str] = {"none", "low", "medium", "high"}
 TLM_VALID_LOG_OPTIONS: Set[str] = {"perplexity", "explanation"}
-TLM_VALID_GET_TRUSTWORTHINESS_SCORE_KWARGS: Set[str] = {"perplexity"}
+TLM_VALID_GET_TRUSTWORTHINESS_SCORE_KWARGS: Set[str] = {"perplexity", "constrain_outputs"}
 TLM_VALID_KWARGS: Set[str] = {"constrain_outputs"}
```
huiwengoh, line 40: [nit] I think would be clearer to make this more explicit distinction?

```suggestion
TLM_VALID_GET_TRUSTWORTHINESS_SCORE_KWARGS: Set[str] = {"perplexity", "constrain_outputs"}
TLM_VALID_PROMPT_KWARGS: Set[str] = {"constrain_outputs"}
```

cleanlab_studio/internal/tlm/validation.py
```diff
@@ -23,16 +23,18 @@
...
-        )
+    if valid_kwargs_check:
+        # validate kwargs - only allow constrain_outputs
+        supported_kwargs = TLM_VALID_KWARGS
+        unsupported_kwargs = set(kwargs.keys()) - supported_kwargs
+        if unsupported_kwargs:
+            raise ValidationError(
+                f"Unsupported keyword arguments: {unsupported_kwargs}. "
+                f"Supported keyword arguments are: {supported_kwargs}"
+            )
```
huiwengoh, line 37: I think we should separate this check helper function into a few separate pieces - one to check if the kwargs are valid, and then another one to specifically validate what users pass into `constrain_outputs`. That second function can then be reused for both `prompt` and `score` in a cleaner way

cleanlab_studio/internal/tlm/validation.py
```diff
@@ -263,10 +265,29 @@ def validate_tlm_options(options: Any) -> None:
...
+            if response not in constrain_outputs:
+                raise ValidationError(
+                    f"Response '{response}' must be one of the constraint outputs: {constrain_outputs}"
+                )
+        else:
+            for i, (resp, constraints) in enumerate(zip(response, constrain_outputs)):
+                if constraints is not None and resp not in constraints:
+                    raise ValidationError(
+                        f"Response '{resp}' at index {i} must be one of the constraint outputs: {constraints}"
+                    )
```
huiwengoh, line 289: I think this section can be nested under the `if not SKIP_VALIDATE_TLM_OPTIONS:` clause since it is validation (mostly for internal testing/debugging purposes if we need)

jas2600: refactored the existing and new functions a bit, and now it's calling `process_and_validate_kwargs_constrain_outputs` which includes both validation and process the constrain_outputs. i don't see a strong use case to skip the validation for constrain_outputs since we will have the same check from backend even if we skip here. To test the validation logic from backend API, we can test through the direct API call. So will keep this outside of `if not SKIP_VALIDATE_TLM_OPTIONS:`

jas2600:
will make changes directly to the new python client actually

